PR #141, opened 2025-04-24
Add References to Hardware spec:

jacobweinstock:
## Description

This allows users to define CRDs and link and reference them in a Tinkerbell Hardware object. This makes the reference data available to Templates. Users have routinely asked for integrations into other data sources, Netbox for example. This provides one avenue for integration without needing in-tree code to handle all the different data sources.

Allow and deny lists are required to make objects available for lookup. By default, all objects are denied. Tink Controller CLI flags/Env vars must be used to allow access to objects in the cluster.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


rpardini (approved):

crd/bases/tinkerbell.org_hardware.yaml
```diff
@@ -355,6 +355,35 @@ spec:
                   state:
                     type: string
                 type: object
+              references:
+                additionalProperties:
+                  properties:
+                    group:
```
displague, line 361: Is this apiGroupVersion? A description would help here. 

This field is the dividing line between the well documented fields below and the undocumented fields above.

docs/REFERENCES.md
```diff
@@ -0,0 +1,94 @@
+# Hardware References
+
+This doc will explain what Hardware references are, how to define them, and how to use them in a Template.
+
+## What are References?
+
+The Hardware custom resource defines a field, `spec.references`, which allows for 
+
+The string name under `spec.reference` can be anything and will be used to reference the object in the Template. The name is not required to be the same as the name of the object being referenced.
```
displague: which allows for...
(And spec.reference is not a thing)

which provides a list of resources types that can be accessed to retrieve Hardware data. Each reference, must have a string name. It can be anything ...

docs/REFERENCES.md
```diff
@@ -0,0 +1,94 @@
...
+
+```yaml
+spec:
+  references:
+    lvm:
+      group: example.org
+      name: lvm1
+      namespace: tink
+      resource: lvms
+      version: v1
```
displague: I didn't see version in the definition and assumed it was include in group.

PR #142, opened 2025-04-25
Add CAPT

jacobweinstock:
## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
Walking down this path to bring in CAPT, I've decided that this isn't something I want to do. At least not right now.

PR #150, opened 2025-04-28
tootles: introduce `/tootles/token/:token/2009-04-04` metadata prefix endpoint

rpardini:
#### tootles: introduce `/tootles/token/:token/2009-04-04` metadata prefix endpoint

- using Hardware CRD's .Metadata.Token, one can now get info out of tootles without needing a stable IP address
- in cloud-init.cfg: `metadata_urls: ["http://<tootlesHost>:50061/tootles/token/<hwToken>"]`
- introduce `GetEC2InstanceByMetadataToken()` member for Client interface
    - add indexed lookup via `.Spec.Metadata.Token` for kube backend's impl via `hwByMetadataToken()`
    - no-op impl for noop backend
- frontend: extract method `writeInstanceDataOrErrToHTTP()`
  - call it from both IP-based and Token-based route groupings
- some maintenance tasks missing from Makefile:
    - ran gofumpt
      - `go install mvdan.cc/gofumpt@latest`
      - `gofumpt -l -w .`
    - updated tootles' frontend_mock_test.go
      - `go install go.uber.org/mock/mockgen@latest`
      - `cd tootles/internal/frontend/ec2`
      - `mockgen -source=frontend.go > frontend_mock_test.go`
      - then manually fix pkg change etc? I guess I'm missing something from old times

#### api: Hardware: add .Spec.Metadata.Token string field

- this is meant as an alternative to Hardware lookups by IP or by MAC/ID


jacobweinstock:
Hey @rpardini , thanks for this. Would you be open to using an existing field in the Hardware spec? `spec.metadata.instance.id` for example. This value in cloud-init is supposed to be unique. [ref](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#v1-instance-id).

rpardini:
Indeed - but exposing this new endpoint, allowing for anyone who knows the Metadata.Instance.Id to get its data, might be a bit too revealing. We know users shouldn't have secrets in user-data, for example, but most CAPI providers do. 

I added the separate and unrelated Token field so that users would have to be explicit about exposing it; ideally I should have schema-limited it to avoid slashes, too, as it is an url component here.

rpardini:
I will rework this to use `spec.metadata.instance.id`, maybe in combination with a configured "salt" and hashing for added security.

Issue #159, opened 2025-05-07
Duplicate Service type


matiaslarson:
The `tinkerbell` Service spec defines both a hardcoded value and and a variable read from the values.yaml file leading to an error:
https://github.com/tinkerbell/tinkerbell/blob/main/helm/tinkerbell/templates/service.yaml#L9-L10
```kubectl apply -k .
error: map[string]interface {}(nil): yaml: unmarshal errors:
  line 11: mapping key "type" already defined at line 10```

PR #160, opened 2025-05-07
fix: remove duplicate `type`

matiaslarson:
## Description

This PR removes the duplicate `type` from the Service definition
## Why is this needed

Fixes: #159 

Fixes a bug, no additional steps are required.

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock (approved):
Ah good catch, thank you for this. @matiaslarson

PR #202, opened 2025-06-12
add hostname to auto ipxe script

yogev-lb:
when booting osie we need to know the hostname to update the metadata and hw-info on self-test.

today we don't have any way to pass on a single hardware information that would be passed to the osie when first boot.

this extention will add the id={hostname} to the /proc/cmdline which will allow hwinfo process to extract this field and update selftest.

on v1beta1 we will have a more generic way to pass in kernel args on the hardware level.

we used to use facility but for some-reason this is not passed on by https://github.com/tinkerbell/dhcp which is the current kube-backend for tink:

When the OSIE boots our tink-agent is using the hostname to lookup some values and configure our metadata server. We need this info when we boot, before we start the rootfs inauguration.

https://github.com/tinkerbell/dhcp/blame/d8def67df9bdce1ff5269cd475094f08f25900bf/backend/kube/kube.go#L276

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/smee/internal/ipxe/script/hook.go
+++ b/smee/internal/ipxe/script/hook.go
@@ -18,7 +18,7 @@ set retry_delay:int32 {{ .RetryDelay }}
 set idx:int32 0
 :retry_kernel
 kernel ${download-url}/${kernel} {{- if ne .VLANID "" }} vlan_id={{ .VLANID }} {{- end }} {{- range .ExtraKernelParams}} {{.}} {{- end}} \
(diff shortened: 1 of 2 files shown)


jacobweinstock:
Hey @yogev-lb , thanks for opening this PR. This repo allows using the `spec.metadata.facility.facility_code` to add arbitrary kernel parameters.

ref: https://github.com/tinkerbell/tinkerbell/blob/b85407b4f0fae7690b30823c6e58b684846c413a/pkg/backend/kube/smee.go#L229

example:
```yaml
spec:
  metadata:
    facility:
      facility_code: onprem id=123abc
```

PR #224, opened 2025-06-26
Add port to HTTP filename in DHCP examples

mmeier86:
By default, the iPXE script is delivered by an HTTP server listening on port 7171, not the default port 80, so the port needs to be provided in the bootfile name send to the host.

## Description

A very small docs fix. In the examples on how to configure DHCP servers for delivering the iPXE script, the HTTP URLs don't contain the default `7171` port used when deploying Tinkerbell via the Helm chart, meaning hosts trying to use those examples would not be able to fetch the file.

## Why is this needed

I've not raised an issue for this.

Fixes: #

## How Has This Been Tested?
This is a docs change, revealed to me by hammering onto a dnsmasq config and staring at tcpdump output for a long long time before realising my DHCP setup was fine, I was just missing a port in the HTTP URL. 

No changes required from users. Impact should be a lot of time saved staring at tcpdumps trying to figure out what's wrong.

## Checklist:

I have:

### DIFF
--- a/docs/technical/smee/DHCP.md
+++ b/docs/technical/smee/DHCP.md
@@ -40,6 +40,9 @@ Most DHCP services all customization of a `next server` option. This option gene
 
 The following code snippets are generic examples of the config needed to enable the two-step process to an existing DHCP service. It does not cover the IPAM info that is also required.
 
+**NOTE:** The HTTP URL pointing to the `auto.ipxe` script may need a port defined, depending on how Tinkerbell is deployed.
+          At the time of writing, the Tinkerbell Helm chart deploys the Smee HTTP server responsible for delivering the iPXE script on port `7171`.
+
 [dnsmasq](https://linux.die.net/man/8/dnsmasq)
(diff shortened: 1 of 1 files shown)


jacobweinstock (approved):
Hey @mmeier86, thanks for this. Apologies that you had to figure this out the long way. I ported these docs over from github.com/tinkerbell/smee and didn't properly review them.

mmeier86:
@jacobweinstock Of course, I was just reading up on that after I saw the failure. Will do so now and provide a fresh push in a a bit.

> Hey @mmeier86, thanks for this. Apologies that you had to figure this out the long way. I ported these docs over from github.com/tinkerbell/smee and didn't properly review them.

No need to apologise. :-) I stared at that config line for a long time and somehow still missed that it didn't have a port set. It's mostly on me for not connecting a keyboard and just going into iPXE to get a better look at the error it was showing only briefly before the Pi's UEFI menu comes up.

Issue #261, opened 2025-07-19
Broadcast-interface init container's "-s" option seems to be broken


mmeier86:
I'm currently working on updating the playground to the `v0.19.1` release, and netbooting/provisioning the new VM consistently fails because there are no DHCP replies.

This seems to be due to the script in [host-interface-config-map.yaml](https://github.com/tinkerbell/tinkerbell/blob/v0.19.1/helm/tinkerbell/templates/host-interface-config-map.yaml) not handling the `-s eth1` option correctly. As a consequence, when running the playground with the libvirt provider, the DHCP broadcast interface created by the Helm chart is connected to the [management network/interface](https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#management-network) the libvirt provider creates, not the network created by the Vagrantfile.

Here's an example output for the `broadcast-interface` init container:
```bash
+ src_interface=                                                                                                       
+ interface_type=macvlan                                                                                               
+ interface_mode=bridge                                                                                                
+ clean=false                                                                                                                                                                                                                                 
+ getopt -a -o s::t:ch --long src::,type:,clean,help -- -s eth1 -t macvlan                                             
+ args=' -s '"''"' -t '"'"'macvlan'"'"' -- '"'"'eth1'"'"                                                               
+ '[[' 0 -gt 0 ]]                                                                                                      
+ eval set -- -s ''"''" -t ''"'"'macvlan'"'" -- ''"'"'eth1'"'"                                                         
+ set -- -s  -t macvlan -- eth1                                                                                        
+ :                                                                                                                                                                                                                                           
+ '[['  '==' -- '||'  '==' '-*' ]]                                                                                     
+ src_interface=                                                                                                       
+ shift 2                                                                                                                                                                                                                                     
+ :                                                                                                                    
+ '[[' macvlan '==' ipvlan ]]                                                                                          
+ shift 2                                                                                                              
+ :                                                                                                                    
+ shift                                                                                                                
+ break                                                                                                                
+ '[[' -z  ]]                                                                                                          
+ nsenter -t1 -n ip route+ head -n1                                                                                    
                                                                                                                       
+ awk '/default/ {print $5}'                                                                                           
+ src_interface=eth0                                                                                                   
+ false                                                                                                                
+ main eth0 macvlan bridge                                                                                             
+ local 'src_interface=eth0'                                                                                           
+ local 'interface_type=macvlan'                                                                                       
+ local 'interface_mode=bridge'                                                                                        
[...]

PR #262, opened 2025-07-19
Fix broadcast-interface container command

mmeier86:
## Description

This PR fixes an issue with the `-s` option on the `host-interface-config-map.yaml` script introduced in f950881513e30453d4f68d8ebff6f80231024814. In that commit, the `-s` argument was made optional. As per the [getopt man page](https://www.man7.org/linux/man-pages/man1/getopt.1.html), options with optional arguments have to have the argument following the option directly, without spaces.

Due to this, the `-s` option as given in the [broadcast-interface container's command](https://github.com/tinkerbell/tinkerbell/blob/863c916e14326fd446c414637995b7388494080b/helm/tinkerbell/templates/deployment.yml#L319) leads to the argument always being ignored and the default approach of using the interface with the default route being used instead.

## Why is this needed

Fixes: #261 

## How Has This Been Tested?
I ran the playground with the libvirt provider and:
- logged into the `stack` VM
- Edited the deployment with `kubectl edit -n tink-system deployment.apps/tinkerbell`
- Joined the `-s` parameter to read `-seth1`, as this PR will do
- Ran through a full provisioning as per the playground setup
- Also verified that the init container used `eth1` now by looking at its logs

There is a potential effect on users who set `deployment.init.sourceInterface` and never realized that it wasn't working. Their setup might suddenly break if e.g. the interface they provided doesn't exist or isn't correctly configured.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/helm/tinkerbell/templates/deployment.yml
+++ b/helm/tinkerbell/templates/deployment.yml
@@ -316,7 +316,7 @@ spec:
       initContainers:
       - name: broadcast-interface
         image: {{ .Values.deployment.init.image }}
-        command: ["/script/host_interface.sh", "-s", "{{ $sourceInterface }}", "-t", "{{ $dhcpInterfaceType }}"]
+        command: ["/script/host_interface.sh", "-s{{ $sourceInterface }}", "-t", "{{ $dhcpInterfaceType }}"]
         volumeMounts:
             - name: script
               mountPath: "/script"


jacobweinstock:
Hey @mmeier86. Thanks for reporting and PRing a fix for this. I hope you don't mind, I'd like to fix the script to behave properly instead of fixing the invocation in the helm chart. I've opened https://github.com/tinkerbell/tinkerbell/pull/263 to do this.

mmeier86:
Yeah, that's definitely the better approach. I will close this one then.

PR #321, opened 2025-08-12
Tink Agent: filter out block devices with unknown controller type

kmullin:
## Description

while testing inside VMs, the auto-discovery feature of the tink-agent discovers disks /dev/ram{1..15} and /dev/nbd{0..15}. This change filters those out and prefers block devices which have a controllerType which isnt Loop or Unknown

This was found on HookOS 0.10.0-build-196.

## Why is this needed

I'm unsure if *all* block devices are going to be discovered 

Fixes: #

## How Has This Been Tested?

At least in my lab, on VMs and limited hardware types, this still correctly detects block devices which have an actual vendor reported from the device, and does not add every single "block" device to `spec.disks` in the generated `Hardware` object.

~The underlying `ghw` package reads from `/sys/block/<dev>/device/vendor`, and if the file doesn't exist will default to `"unknown"`~ [seen here](https://github.com/jaypipes/ghw/blob/v0.17.0/pkg/block/block_linux.go#L82-L91).

I am fairly confident this is safe, but due to limited hardware availability, I'd love to get some feedback.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/tink/agent/internal/attribute/attribute.go
+++ b/tink/agent/internal/attribute/attribute.go
@@ -88,7 +88,7 @@ func DiscoverBlockDevices(l logr.Logger) []*data.Block {
 		if d == nil {
 			continue
 		}
-		if d.StorageController != block.STORAGE_CONTROLLER_LOOP {
+		if d.StorageController != block.StorageControllerLoop && d.StorageController != block.StorageControllerUnknown {
 			blockDevices = append(blockDevices, &data.Block{
 				Name:              toPtr(d.Name),
 				ControllerType:    toPtr(d.StorageController.String()),


tink/agent/internal/attribute/attribute.go
```diff
@@ -85,7 +85,8 @@ func DiscoverBlockDevices(l logr.Logger) []*data.Block {
 	}
 	var blockDevices []*data.Block
 	for _, d := range b.Disks {
-		if d == nil {
+		// XXX: inside a VM, we see /dev/ramXX and the underlying library seems to think its a BlockDevice
+		if d == nil || d.Vendor == "unknown" {
```
jacobweinstock: Hey @kmullin, thanks for this. We'll need a different approach here though. With this logic, the following devices would be excluded. 

This is from a physical machine I have:
```json
{
    "name": "nvme0n1",
    "controllerType": "NVMe",
    "driveType": "SSD",
    "size": "239GB",
    "physicalBlockSize": "512B",
    "vendor": "unknown",
    "model": "KINGSTON OM8PDP3256B-A01"
}
```
This is from a `qemu` VM I regularly use to test:
```json
{
    "controllerType": "NVMe",
    "driveType": "SSD",
    "model": "unknown",
    "name": "nvme0n1",
    "physicalBlockSize": "512B",
    "size": "10GB",
    "vendor": "unknown"
},
```

Maybe one of the following options could be used.

```Go
if d == nil || strings.HasPrefix(d.Name, "nbd") || strings.HasPrefix(d.Name, "ram") {
	continue
}
```
OR
```Go
if d.StorageController != block.STORAGE_CONTROLLER_LOOP && d.StorageController != block.STORAGE_CONTROLLER_UNKNOWN {
...
}
```

I'm leaning toward the second one. What are your thoughts?

jacobweinstock: Here are 2 more entries from a Raspberry PI 5. The SD card and the NVMe drive show an `unknown` vendor.

```json
{
    "controllerType": "MMC",
    "driveType": "SSD",
    "model": "unknown",
    "name": "mmcblk0",
    "physicalBlockSize": "512B",
    "size": "29GB",
    "vendor": "unknown"
}
```

```json
{
    "controllerType": "NVMe",
    "driveType": "SSD",
    "model": "unknown",
    "name": "nvme0n1",
    "physicalBlockSize": "512B",
    "size": "233GB",
    "vendor": "unknown"
}
```

kmullin: You're amazing! Thanks for double checking this. Yes I suppose my solution is not great.

I'll take a closer look at what you've proposed with the second option and test out on some other hardware too!

kmullin: yeah I see that the controller type is basically set from device prefixes already: https://github.com/jaypipes/ghw/blob/e75a40d23a6cb2e7314ccedd4a20a2e92542332f/pkg/block/block_linux.go#L382-L420

So that is likely better than doing another prefix filtering.

I've also updated the references to use the non-deprecated versions of the same thing.

This appears to work in my testing environment the same way.

jacobweinstock (approved):
Thanks again for this!

jacobweinstock:
> Sure thing! I force pushed a signed commit. Sorry for the troubles.

No worries at all, thanks for doing it!

PR #343, opened 2025-08-29
add nodeselector to kubevip

sepauli:
## Description
We use a special network setup, which is why only certain nodes can perform a PXE boot.
We have labeled the nodes so that Tinkerbell resources, only start on these nodes.
By default, kube-vip is started on all control planes, which means that Tinkerbell does not work properly.
We have determined that kube-vip must be started only on the nodes that are authorized to perform a PXE boot
Therefore, we would like to add a node selector to kube-vip to control that kube-vip only runs on the authorized nodes

## How Has This Been Tested?
Deployed the stack with a given nodeSelector


jacobweinstock (approved):
Thanks for this, @sepauli

jacobweinstock:
> This pull request has been removed from the queue for the following reason: `pull request dequeued`.
> 
> Pull request #343 has been dequeued. The pull request could not be merged. This could be related to an activated [branch protection](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches) or [ruleset rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) that prevents us from merging. (details: Repository rule violations found
> 
> Commits must have verified signatures.
> 
> ).
> 
> You should look at the reason for the failure and decide if the pull request needs to be fixed or if you want to requeue it. If you do update this pull request, it will automatically be requeued once the queue conditions match again. If you think this was a flaky issue instead, you can requeue the pull request, without updating it, by posting a `@mergifyio requeue` comment.

Hey @sepauli, apologies. Would you mind setting up the GitHub verified signatures?

PR #358, opened 2025-09-05
Add TLS termination

jacobweinstock:
## Description

This adds TLS termination. This also allows URL patterns to be served from both HTTP and HTTPS. HTTP and HTTPS will be served on different ports. There is a doc in this PR that describes the details.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


s3rj1k:
> What's the motivation for wanting to replace nginx?

Basically reducing SBOM, so people won't need to worry about nginx CVEs

PR #360, opened 2025-09-09
helm: no service when using hostNetwork

clwluvw:
## Description

There is no need to deploy the service when we have the deployment running with `hostNetwork: true`. This just makes it confusing.


jacobweinstock (approved):
Thanks, @clwluvw !

PR #372, opened 2025-09-13
dhcp: add RFC 3442 classless static routes support

clwluvw:
## Description

Implement RFC 3442 classless static routes for DHCP option 121.

## How Has This Been Tested?

Configure the `classless_static_routes` config in the hardware spec and expect option 121 in the DHCP response.

Default is empty, so no impact.

## Checklist:

I have:

- [x] added unit or e2e tests


smee/internal/dhcp/handler/reservation/option.go
```diff
@@ -50,6 +50,13 @@ func (h *Handler) setDHCPOpts(_ context.Context, _ *dhcpv4.DHCPv4, d *data.DHCP)
 	if h.SyslogAddr.Compare(netip.Addr{}) != 0 {
 		mods = append(mods, dhcpv4.WithOption(dhcpv4.OptGeneric(dhcpv4.OptionLogServer, h.SyslogAddr.AsSlice())))
 	}
+	if len(d.ClasslessStaticRoutes) > 0 {
+		if routeBytes, err := encodeClasslessStaticRoutes(d.ClasslessStaticRoutes); err == nil {
+			mods = append(mods, dhcpv4.WithGeneric(dhcpv4.OptionClasslessStaticRoute, routeBytes))
```
jacobweinstock: FYI, there is a [`dhcpv4.OptClasslessStaticRoute`](https://github.com/insomniacslk/dhcp/blob/d3abe7ccb0ad34d401030e9e1365c83075732563/dhcpv4/option_routes.go#L99) function that could be useful here.

smee/internal/dhcp/handler/reservation/option_test.go
```diff
@@ -362,3 +408,125 @@ func TestSetNetworkBootOpts(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := encodeClasslessStaticRoutes(tt.routes)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("encodeClasslessStaticRoutes() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !tt.wantErr && !bytesEqual(got, tt.want) {
```
jacobweinstock: Can `bytes.Equal` or `bytes.EqualFold` be used here instead?

smee/internal/dhcp/handler/reservation/option.go
```diff
@@ -123,3 +130,46 @@ func (h *Handler) bootfileAndNextServer(ctx context.Context, pkt *dhcpv4.DHCPv4,
...
 }
+
+// encodeClasslessStaticRoutes encodes static routes according to RFC 3442 format.
+// RFC 3442 defines the format for DHCP option 121 (Classless Static Route Option).
+// The format is: width | destination descriptor | router
+// where:
+// - width: number of significant bits in the subnet mask (0-32)
+// - destination descriptor: variable length based on width
+// - router: 4 bytes for IPv4 address of the router
+func encodeClasslessStaticRoutes(routes []data.StaticRoute) ([]byte, error) {
```
jacobweinstock: FYI, the https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/option_routes.go  package provides this same functionality. It has a [`Route`](https://github.com/insomniacslk/dhcp/blob/d3abe7ccb0ad34d401030e9e1365c83075732563/dhcpv4/option_routes.go#L12) struct and a [`ToBytes`](https://github.com/insomniacslk/dhcp/blob/d3abe7ccb0ad34d401030e9e1365c83075732563/dhcpv4/option_routes.go#L78) function.

api/v1alpha1/tinkerbell/hardware.go
```diff
@@ -193,6 +196,17 @@ type IP struct {
...
 
+// ClasslessStaticRoute represents a classless static route for DHCP option 121 (RFC 3442).
+type ClasslessStaticRoute struct {
+	// DestinationDescriptor is the network address and prefix length.
+	// The format is "network/prefixlength", e.g., "192.168.1.0/24" or "10.0.0.0/8".
+	// +kubebuilder:validation:Pattern=`^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/(3[0-2]|[12]?[0-9])$`
+	DestinationDescriptor string `json:"destination_descriptor,omitempty"`
+	// Router is the IP address of the router for this route.
+	// +kubebuilder:validation:Pattern=`^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`
+	Router string `json:"router,omitempty"`
```
jacobweinstock: Do these kubebuilder validation patterns prohibit an empty string from being provided? Is there a way to require both fields to be provided if any of them are a non-empty string?

smee/internal/dhcp/handler/reservation/option.go
clwluvw: Thanks for the hint, I have replaced it with that struct completely.

smee/internal/dhcp/handler/reservation/option_test.go
clwluvw: it's gone now.

api/v1alpha1/tinkerbell/hardware.go
clwluvw: i have removed `omitempty`, so now in the CRD you can see they are both required. Also, with the validation of the regex we have, if you provide an empty string, then it will fail due to not matching the regex.

jacobweinstock (approved):
Thanks, @clwluvw!

clwluvw:
@jacobweinstock, thank you for the quick review. Do you have any date for the next release that could include this?

jacobweinstock:
@clwluvw, I'm going to cut a new release now, actually. The things I wanted to land need some more time.

PR #383, opened 2025-09-22
dhcp: allow dhcp to reply custom TFTPServerName and BootFileName

clwluvw:
## Description

This change introduces two new optional fields to the DHCP spec:

- TFTPServerName: identifies the TFTP/boot server host or address
- BootFileName: identifies the boot file to request from the server

These are required by some PXE/iPXE/ONIE clients for network boot, where they require their own config and img and not respecting ipxe scripts.

A validation rule has been added to enforce that both fields must be specified together, or both left empty. This rule is applied both via kubebuilder XValidation in the Go type definition and in the CRD schema (x-kubernetes-validations).

## How Has This Been Tested?

Configure both fields see them in response.

No impact, all are options.

## Checklist:

I have:

- [x] added unit or e2e tests


jacobweinstock:
Hey @clwluvw. Thanks for this. For these 3 special types of netboot clients (PXE/iPXE/ONIE), mind elaborating on the current issues with Tinkerbell? I'm also interested in understanding more about this statement, mind expanding on this? "they require their own config and img and not respecting ipxe scripts."

smee/internal/dhcp/handler/reservation/handler.go
```diff
@@ -198,9 +198,13 @@ func (h *Handler) updateMsg(ctx context.Context, pkt *dhcpv4.DHCPv4, d *data.DHC
 		dhcpv4.WithGeneric(dhcpv4.OptionServerIdentifier, h.IPAddr.AsSlice()),
 		dhcpv4.WithServerIP(h.IPAddr.AsSlice()),
 	}
+
 	mods = append(mods, h.setDHCPOpts(ctx, pkt, d)...)
 
-	if h.Netboot.Enabled && dhcp.IsNetbootClient(pkt) == nil {
+	// Only do auto-generation if BOTH TFTPServerName and BootFileName are empty.
```
jacobweinstock: what is "auto-generation"?

api/v1alpha1/tinkerbell/hardware.go
```diff
@@ -187,6 +188,16 @@ type DHCP struct {
...
+	// TFTPServerName is the TFTP server name or IP address.
+	// This is used by some network boot clients to locate the TFTP server.
+	// If specified, BootFileName must also be specified.
+	//+optional
+	TFTPServerName string `json:"tftp_server_name,omitempty"`
+	// BootFileName is the boot file name.
+	// This is used by some network boot clients to locate the boot file.
+	// If specified, TFTPServerName must also be specified.
+	//+optional
+	BootFileName string `json:"bootfile_name,omitempty"`
```
jacobweinstock: Please add some code comments detailing why these exist and when they should/can be used. NVOS, etc...

jacobweinstock (changes requested):
Thanks for this, @clwluvw. I left a couple comments.

smee/internal/dhcp/handler/reservation/handler.go
clwluvw: I realize that was poorly worded; I meant the case where Tinkerbell will generate those fields. I have now refined the comment.

api/v1alpha1/tinkerbell/hardware.go
clwluvw: done

api/v1alpha1/tinkerbell/hardware.go
```diff
@@ -187,6 +188,18 @@ type DHCP struct {
...
+	// like NVIDIA NVOS switches.
+	// If specified, BootFileName must also be specified.
+	//+optional
+	TFTPServerName string `json:"tftp_server_name,omitempty"`
+	// BootFileName is the boot file name (DHCP option 67).
+	// Used for explicit boot file configuration, required by some network boot clients
+	// like NVIDIA NVOS switches.
+	// If specified, TFTPServerName must also be specified.
+	//+optional
+	BootFileName string `json:"bootfile_name,omitempty"`
```
jacobweinstock: should the json be `boot_file_name`?

smee/internal/dhcp/handler/reservation/handler_test.go
```diff
@@ -959,3 +963,134 @@ func TestEncodeToAttributes(t *testing.T) {
 		})
 	}
 }
+
+// TestAutoGenerationReflection tests the key behavior: when TFTPServerName or BootFileName are set,
+// auto-generation should be completely disabled
+func TestAutoGenerationReflection(t *testing.T) {
+	tests := map[string]struct {
+		name           string
```
jacobweinstock: Doesn't look like this field is used?

smee/internal/dhcp/handler/reservation/handler.go
jacobweinstock: thank you

api/v1alpha1/tinkerbell/hardware.go
clwluvw: done. i was trying to follow the dhcp option naming schema.

smee/internal/dhcp/handler/reservation/handler_test.go
clwluvw: right, i refactored the test and forgot to drop the field. thanks. I have also renamed the test to reflect that fact to drop the auto-generation phrase.

api/v1alpha1/tinkerbell/hardware.go
jacobweinstock: That's fair. The [doc](https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.xhtml#:~:text=67,Boot%20File%20Name) is confusing too as it seems to have it both ways. Thanks for updating it tho!

api/v1alpha1/tinkerbell/hardware.go
```diff
@@ -169,6 +169,7 @@ type OSIE struct {
 }
 
 // DHCP configuration.
+// +kubebuilder:validation:XValidation:rule=(has(self.tftp_server_name) && self.tftp_server_name != "") == (has(self.bootfile_name) && self.bootfile_name != ""),message="TFTPServerName and BootFileName must both be specified or both be empty"
```
jacobweinstock: do these need to be updated? `self.bootfile_name` -> `self.boot_file_name`?

clwluvw: ops, sorry. i tried to come up with a test case for this, but apparently it wasn't that easy. I have adopted the right naming now.

jacobweinstock (approved):
Thanks for working through this with me!

PR #390, opened 2025-09-26
Helm: hookos hostnetwork should be configurable like tinkerbell

jakerbeck:
## Description

Tinkerbell can be configured to use hostNetwork in cases where a load-balancer isn't available, but hookos does not have this ability.  Nodeport could work, but would require the user to manually configure the service port number in values.yaml and then update the artifactsFileServer to use the same port.  This will simplify things for users without load-balancers because they can re-use the `7173` port.

Fixes: #

## How Has This Been Tested?

For my use case, I'm doing some testing with a single node cluster and a qemu VM similar to the metal3 quick start guide (https://book.metal3.io/quick-start).  I was able to test this by patching hookos `kubectl edit deploy hookos` and adding the `hostNetwork: true` configuration.  After doing this I was able to install an OS no the vm.  Wanted to share this solution so others can use it too.

No expected impact, the default will be `hostNetwork: false`

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/helm/tinkerbell/templates/hookos/deployment.yaml
+++ b/helm/tinkerbell/templates/hookos/deployment.yaml
@@ -63,6 +63,7 @@ spec:
         configMap:
           defaultMode: 0700
           name: download-hook
+      hostNetwork: {{ .Values.optional.hookos.hostNetwork }}
       {{- with .Values.optional.hookos.nodeSelector }}
       nodeSelector:
       {{ toYaml . | nindent 8 }}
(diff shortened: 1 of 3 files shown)


jacobweinstock (approved):
Thanks for this contribution, @jakerbeck!

jacobweinstock:
> This pull request has been removed from the queue for the following reason: `pull request dequeued`.
> 
> Pull request #390 has been dequeued. The pull request could not be merged. This could be related to an activated [branch protection](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches) or [ruleset rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) that prevents us from merging. (details: Repository rule violations found
> 
> Commits must have verified signatures.
> 
> ).
> 
> You should look at the reason for the failure and decide if the pull request needs to be fixed or if you want to requeue it. If you do update this pull request, it will automatically be requeued once the queue conditions match again. If you think this was a flaky issue instead, you can requeue the pull request, without updating it, by posting a `@mergifyio requeue` comment.

Hey @jakerbeck, mind setting up your GitHub verified signature please?

PR #395, opened 2025-10-05
Add nocloud datasource

elauriault:
## Description

  Adds NoCloud metadata service to tootles for cloud-init network bonding support on bare metal. Generates network-config from Hardware CRD that cloud-init can use to configure bonded interfaces with static IPs during provisioning.

https://docs.cloud-init.io/en/24.2/reference/datasources/nocloud.html

  Includes:
  - NoCloud frontend (`/meta-data`, `/user-data`, `/network-config` endpoints)
  - Support for all bonding modes (0-6: balance-rr, active-backup, balance-xor, broadcast, 802.3ad, balance-tlb, balance-alb)
  - Correct parameter naming for cloud-init 24.2+ (underscores: `bond-lacp_rate`, `bond-xmit_hash_policy`, etc.)
  - IPv4/IPv6 static IP configuration
  - Backend implementation for kube and noop
  - Tests and documentation

  Fixes: #

  ## How Has This Been Tested?

  - Unit tests pass (`make test`, `make lint`)
  - Integration tested on Rocky Linux 9.6, Ubuntu 22.04, and Ubuntu 24.04
  - Verified schema validation with `cloud-init schema --system`
  - Tested 802.3ad mode with static IPv4 on bare metal with LACP-enabled switch
  - Documented OS-specific issues in `docs/technical/TOOTLES_NOCLOUD.md`:
    - Ubuntu 22.04/24.04 needs `no-nocloud-network.patch` reverted
    - Rocky 9.6 sysconfig renderer needs NetworkManager restart

  No impact. This is opt-in - NoCloud endpoints only respond when requested. EC2 and Hack metadata services unchanged. No migration needed.

  Usage:
  ```yaml
  spec:
    metadata:
      bonding_mode: 4  # 802.3ad
      instance:
        ips:
          - address: "192.168.1.10"
            netmask: "255.255.255.0"
            gateway: "192.168.1.1"
            family: 4

  Kernel parameter: ds=nocloud;seedfrom=http://<tinkerbell-ip>:7172/

  See docs/technical/TOOTLES_NOCLOUD.md for details.

  Checklist:

  I have:

  - updated the documentation and/or roadmap (if required)
  - added unit or e2e tests
  - provided instructions on how to upgrade

### DIFF
--- a/docs/technical/TOOTLES_NOCLOUD.md
+++ b/docs/technical/TOOTLES_NOCLOUD.md
@@ -87,7 +87,7 @@ The NoCloud service is automatically available at the Tinkerbell service endpoin
 
 **PXE kernel parameters:**
 ```
-ds=nocloud;seedfrom=http://<tinkerbell-ip>:7172/
+ds=nocloud;seedfrom=http://<tinkerbell-ip>:7172/nocloud/
 ```
 
 
--- a/docs/technical/TOOTLES_NOCLOUD.md
+++ b/docs/technical/TOOTLES_NOCLOUD.md
@@ -30,23 +30,15 @@ spec:
   interfaces:
     - dhcp:
         mac: b8:cb:29:98:cb:3a
+        name_servers:
+          - "1.1.1.1"
+          - "1.0.0.1"
+          - "2606:4700:4700::1111"
         # First interface
     - dhcp:
         mac: b8:cb:29:98:cb:3b
         # Second interface
 ```
-
-### Supported Bonding Modes
-
-| Mode | Name | Description |
-|------|------|-------------|
-| 0 | balance-rr | Round-robin load balancing |
-| 1 | active-backup | Active-backup fault tolerance |
-| 2 | balance-xor | XOR load balancing |
-| 3 | broadcast | Broadcast on all interfaces |
-| 4 | 802.3ad | IEEE 802.3ad LACP (requires switch support) |
-| 5 | balance-tlb | Adaptive transmit load balancing |
-| 6 | balance-alb | Adaptive load balancing |
 
 ### Generated Network Configuration
 
@@ -79,91 +71,13 @@ config:
       - type: static
         address: 192.168.1.10/24
         gateway: 192.168.1.1
-        dns_nameservers: [8.8.8.8, 8.8.4.4]
+        dns_nameservers: [1.1.1.1, 1.0.0.1]
       - type: static6
         address: 2001:db8::10/64
         gateway: 2001:db8::1
-        dns_nameservers: [2001:4860:4860::8888]
+        dns_nameservers: [2606:4700:4700::1111]
 ```
 
-### Static IP Configuration
-
-#### IPv4 Static IP
-
-Specify IPv4 addresses with netmask:
-
-```yaml
-ips:
-  - address: "192.168.1.10"
-    netmask: "255.255.255.0"    # Converted to /24
-    gateway: "192.168.1.1"
-    family: 4
-```
-
-Supported netmasks:
-- `255.255.255.0` → /24
-- `255.255.255.240` → /28
-- `255.255.255.192` → /26
-- `255.255.0.0` → /16
-- `255.0.0.0` → /8
-
-#### IPv6 Static IP
-
-Specify IPv6 addresses with CIDR prefix:
-
-```yaml
-ips:
-  - address: "2001:db8::10/64"
-    gateway: "2001:db8::1"
-    family: 6
-```
-
-#### Dual-Stack Configuration
-
-Configure both IPv4 and IPv6:
-
-```yaml
-ips:
-  - address: "192.168.1.10"
-    netmask: "255.255.255.0"
-    gateway: "192.168.1.1"
-    family: 4
-  - address: "2001:db8::10/64"
-    gateway: "2001:db8::1"
-    family: 6
-```
-
-### Bond Parameters by Mode
-
-#### Mode 4 (802.3ad LACP)
-
-```yaml
-params:
-  bond-mode: 802.3ad
-  bond-miimon: 100
-  bond-lacp_rate: fast           # LACP packet rate (slow|fast)
-  bond-xmit_hash_policy: layer3+4 # Load balancing hash policy
-  bond-ad_select: stable          # Aggregation selection
-```
-
-#### Mode 1 (active-backup)
-
-```yaml
-params:
-  bond-mode: active-backup
-  bond-miimon: 100
-  bond-primary_reselect: always
-  bond-fail_over_mac: none
-```
-
-#### Mode 5 (balance-tlb)
-
-```yaml
-params:
-  bond-mode: balance-tlb
-  bond-miimon: 100
-  bond-tlb_dynamic_lb: 1
-```
 
 ### Cloud-init Integration
 
@@ -174,14 +88,6 @@ The NoCloud service is automatically available at the Tinkerbell service endpoin
 ds=nocloud;seedfrom=http://<tinkerbell-ip>:7172/
 ```
 
-**Or in cloud-init config:**
-```yaml
-# /etc/cloud/cloud.cfg.d/99-tinkerbell.cfg
-datasource_list: [NoCloud]
-datasource:
-  NoCloud:
-    seedfrom: http://<tinkerbell-ip>:7172/
-```
 
 ## OS-Specific Notes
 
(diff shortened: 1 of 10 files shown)


pkg/data/instance.go
```diff
@@ -42,6 +42,14 @@ type LicenseActivation struct {
 	State string
 }
 
+// NoCloudInstance is a struct that contains the hardware data exposed from the NoCloud API endpoints.
+// It supports network bonding and IPv6 configuration for bare metal servers.
+type NoCloudInstance struct {
+	Userdata      string
+	Metadata      Metadata
+	NetworkConfig interface{} // YAML network configuration data
```
jacobweinstock: Can this be a defined struct? Having a defined spec would improve the understandability, testability, and maintainability. The spec from [here](https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html) could be used to define a Go struct.

tootles/internal/frontend/nocloud/frontend.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+}
+
+// Configure configures router with the supported NoCloud metadata API endpoints.
+func (f Frontend) Configure(router gin.IRouter) {
+	// Configure NoCloud endpoints directly under the root path
+	router.GET("/meta-data", f.metaDataHandler)
+	router.GET("/user-data", f.userDataHandler)
+	router.GET("/vendor-data", f.vendorDataHandler)
+	router.GET("/network-config", f.networkConfigHandler)
```
jacobweinstock: can we prefix these with `nocloud`?

tootles/internal/frontend/nocloud/frontend.go
```diff
@@ -0,0 +1,178 @@
...
+			},
+		}
+
+		yamlData, err := yaml.Marshal(fallbackConfig)
+		if err != nil {
+			_ = ctx.AbortWithError(http.StatusInternalServerError, fmt.Errorf("failed to marshal fallback config: %w", err))
+			return
+		}
+
+		ctx.Header("Content-Type", "text/yaml")
```
jacobweinstock: I couldnt find any docs on what cloud-init expects for the content type. Closest I found was [this](https://cloudinit.readthedocs.io/en/stable/explanation/format.html#headers-and-content-types), but no mention of network-config. Have you found any documentation of this? The other endpoints should probably follow these content types.

jacobweinstock (changes requested):

tootles/internal/frontend/nocloud/frontend.go
elauriault: Fixed in f5ed8df9

pkg/data/instance.go
elauriault: Fixed in f55dec1b

tootles/internal/frontend/nocloud/frontend.go
elauriault: I couldn't find any specific documentation on the content-type either. cloud-init 24.x seems ok with both. We could just use text/plain for consistency.

pkg/data/instance.go
```diff
@@ -42,6 +42,131 @@ type LicenseActivation struct {
 	State string
 }
 
+// NoCloudInstance is a struct that contains the hardware data exposed from the NoCloud API endpoints.
+// It supports network bonding and IPv6 configuration for bare metal servers.
+type NoCloudInstance struct {
+	Userdata      string
+	Metadata      Metadata
+	NetworkConfig interface{} // Network Config data (typically NetworkConfigV2)
+	// Note: Uses interface{} for flexibility in YAML marshaling. The actual structure
```
jacobweinstock: Sorry, im not following here. What flexibility does an `interface{}` provide over using the `NetworkSpec` directly?

elauriault: The interface{} was left  to allow for support of multiple network config versions if we ever needed to. Since we've simplified to only support v2, I switched it to *NetworkConfig (which is a wrapper for NetworkSpecV2) in 61f727ed.

docs/technical/TOOTLES_NOCLOUD.md
```diff
@@ -0,0 +1,108 @@
...
+Configure bonding and static IPs in your Hardware resource:
+
+```yaml
+apiVersion: tinkerbell.org/v1alpha1
+kind: Hardware
+metadata:
+  name: server-001
+spec:
+  metadata:
+    bonding_mode: 4  # 802.3ad LACP
```
jacobweinstock: would you mind documenting all the bonding modes, please? This way users don't need to interogate the code.

pkg/backend/kube/tootles.go
```diff
@@ -110,6 +111,294 @@ func toEC2Instance(hw v1alpha1.Hardware) data.Ec2Instance {
...
+				}
+			}
+			break
+		}
+	}
+
+	return ipv4DNS, ipv6DNS
+}
+
+func cidrFromNetmask(netmask string) string {
```
jacobweinstock: FYI, the standard lib has functions to do this same thing. We do it [here](https://github.com/tinkerbell/tinkerbell/blob/main/tink/controller/internal/workflow/template_funcs.go#L40-L54), as an example. Stdlib functions feel and bit more approachable and understandable.

pkg/backend/kube/tootles.go
```diff
@@ -110,6 +111,294 @@ func toEC2Instance(hw v1alpha1.Hardware) data.Ec2Instance {
...
+// Version 2 is the modern Netplan-compatible format.
+// Only generates configuration for network bonding. For non-bonded interfaces, cloud-init handles default DHCP.
+// Returns nil if no network configuration is needed (non-bonded interfaces use default DHCP).
+func generateNetworkConfigV2(hw v1alpha1.Hardware) *data.NetworkConfig {
+	networkSpec := data.NetworkSpecV2{
+		Version: 2,
+	}
+
+	// Check if bonding is enabled
+	bondingEnabled := hw.Spec.Metadata != nil && hw.Spec.Metadata.BondingMode > 0
```
jacobweinstock: Below there is a bonding mode 0 that equals `balance-rr`. This seems to say bonding mode 0 is disabled?

tootles/internal/frontend/nocloud/example_test.go
```diff
@@ -0,0 +1,112 @@
...
+	_ = &http.Server{
+		Addr:    ":8080",
+		Handler: router,
+	}
+
+	log.Println("NoCloud metadata API server starting on :8080")
+	log.Println("Endpoints available:")
+	log.Println("  GET /meta-data")
+	log.Println("  GET /user-data")
+	log.Println("  GET /network-config")
```
jacobweinstock: should these all have a `/nocloud` prefix?

tootles/internal/frontend/nocloud/frontend.go
```diff
@@ -0,0 +1,195 @@
...
+
+// ErrInstanceNotFound indicates an instance could not be found for the given identifier.
+var ErrInstanceNotFound = errors.New("instance not found")
+
+// Client is a backend for retrieving NoCloud Instance data.
+type Client interface {
+	// GetNoCloudInstance retrieves an Instance associated with ip. If no Instance can be
+	// found, it should return ErrInstanceNotFound.
+	GetNoCloudInstance(_ context.Context, ip string) (data.NoCloudInstance, error)
+}
```
jacobweinstock, line 25: The naming here could be improved. "_By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc._" - [Reference](https://go.dev/doc/effective_go#interface-names).

Interface name could maybe something like `InstanceReader` or `NocloudReader`, etc. With the method name being `ReadInstance` or `ReadNocloud`, etc. [Reference](https://github.com/tinkerbell/tinkerbell/blob/main/docs/DESIGN_PHILOSOPHY.md#:~:text=Interfaces%20are%20active%20nouns%3A%20Reader%2C%20Writer%2C%20JobProcessor), https://go.dev/doc/effective_go#Getters

pkg/backend/kube/tootles.go
```diff
@@ -110,6 +111,294 @@ func toEC2Instance(hw v1alpha1.Hardware) data.Ec2Instance {
...
+			bondConfig.Nameservers = &data.NameserversConfig{
+				Addresses: nameservers,
+			}
+		}
+	} else {
+		// Default to DHCP if no static IPs (IPv4 only)
+		bondConfig.Dhcp4 = true
+	}
+
+	bonds["bond0"] = bondConfig
```
jacobweinstock: looks like this only supports a single bond. This should be included in the documentation.

pkg/backend/kube/tootles.go
```diff
@@ -110,6 +111,294 @@ func toEC2Instance(hw v1alpha1.Hardware) data.Ec2Instance {
...
+// Returns separate ethernets and bonds maps.
+// Requires MAC addresses for all interfaces to enable proper matching.
+func generateBondingConfigurationV2(hw v1alpha1.Hardware) (map[string]data.EthernetConfig, map[string]data.BondConfig) {
+	ethernets := map[string]data.EthernetConfig{}
+	bonds := map[string]data.BondConfig{}
+	bondInterfaces := []string{}
+
+	// Create physical interfaces for bonding (without IP addresses)
+	phyIndex := 0
+	for _, iface := range hw.Spec.Interfaces {
```
jacobweinstock, line 358: If i understand this correctly, then all interfaces defined in the Hardware object will be included in the bond configuration. Is that correct? Is that the expected behavior? If so, I think there needs to be some consideration to interfaces that are defined but should not be included in the bond.

jacobweinstock (changes requested):

pkg/backend/kube/tootles.go
elauriault: Good catch. Fixed in 78855b04

pkg/backend/kube/tootles.go
elauriault: Fixed in be0091b2

docs/technical/TOOTLES_NOCLOUD.md
elauriault: Added docs on bonding in fda77ad0

tootles/internal/frontend/nocloud/example_test.go
elauriault: Fixed in 5034a98a

pkg/backend/kube/tootles.go
elauriault: Added docs on current implementation limitations in fda77ad0.
I couldn't a good way to to specify which interfaces should be included in a bond and which ones to leave out with the current CRD. Using references might allow this flexibility at a later time.

pkg/backend/kube/tootles.go
elauriault: Added docs on current implementation limitations in fda77ad0.
I couldn't a good way to to specify which interfaces should be included in which bond with the current CRD. Using references might allow this flexibility at a later time.

tootles/internal/frontend/nocloud/frontend.go
elauriault: I was following the pattern set with ec2.Client and hack.Client.  We could go with nocloud.InstanceReader but we should probably fix ec2 and hack as well.

jacobweinstock:
Hey @leoleg91 , @rnalrd , thanks for expressing your interest in this. Full NoCloud support is planned for the v1apha2 API. Please check out the proposed spec. We'd appreciate any feedback, thanks!

https://github.com/tinkerbell/tinkerbell/tree/main/api%2Fv1alpha2%2Ftinkerbell

https://github.com/tinkerbell/tinkerbell/tree/main/docs%2Ftechnical%2Fv1alpha2

https://github.com/tinkerbell/tinkerbell/pull/445

PR #411, opened 2025-10-14
Add netmaskToPrefix template function

rahulbabu95:
## Description
Adds netmaskToPrefixLength template function to convert netmask notation (e.g., "255.255.255.0") to prefix (e.g., "24").

The function is available in workflow template contexts and will cause template rendering to fail on error. This template function can be handy when using template to configure static IPAM files using tools such as netplan or network manager.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/tink/controller/internal/workflow/template_funcs.go
+++ b/tink/controller/internal/workflow/template_funcs.go
@@ -2,12 +2,15 @@ package workflow
 
 import (
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 )
 
 // templateFuncs defines the custom functions available to workflow templates.
 var templateFuncs = map[string]interface{}{
-	"formatPartition": formatPartition,
+	"formatPartition":       formatPartition,
+	"netmaskToPrefixLength": netmaskToPrefixLength,
 }
 
 // formatPartition formats a device path with partition for the device type. If it receives an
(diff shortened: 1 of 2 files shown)


tink/controller/internal/workflow/template_funcs.go
```diff
@@ -2,12 +2,15 @@ package workflow
...
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 )
 
 // templateFuncs defines the custom functions available to workflow templates.
 var templateFuncs = map[string]interface{}{
 	"formatPartition": formatPartition,
+	"netmaskToCIDR":   netmaskToCIDR,
```
jacobweinstock: Did you consider `prefix` over `cidr` for naming? cidr notation is generally in this format `w.x.y.z/n`, while the prefix is just the `n`.

tink/controller/internal/workflow/template_funcs.go
```diff
@@ -30,3 +33,21 @@ func formatPartition(dev string, partition int) string {
 	}
 	return dev
 }
+
+// netmaskToCIDR converts a netmask (e.g. 255.255.255.0) to CIDR notation (e.g. 24)
+func netmaskToCIDR(netmask string) string {
+	// Parse the netmask
+	ip := net.ParseIP(netmask)
+	if ip == nil {
+		return "24" // Default to /24 if parsing fails
```
jacobweinstock: The defaulting like this means that errors won't cause a Template to fail to render. This means that invalid subnet masks would most likely have to be troubleshooted when the OS comes up. I think a failure to render the Template, which causes the Workflow to not run, might be an improved user experience. Thoughts?

tink/controller/internal/workflow/template_funcs.go
```diff
@@ -30,3 +33,21 @@ func formatPartition(dev string, partition int) string {
 	}
 	return dev
 }
+
+// netmaskToCIDR converts a netmask (e.g. 255.255.255.0) to CIDR notation (e.g. 24)
+func netmaskToCIDR(netmask string) string {
```
jacobweinstock: One option here would be to also take in a default prefix and return a string and an error. This way there can still be defaulting if the user sets a default.

`func nToP(netmask, defaultPrefix string) (string, error) {}`

jacobweinstock (changes requested):
Hey @rahulbabu95. Thanks for this. Couple of comments. Mind adding tests for this function too, please and thank you!

tink/controller/internal/workflow/template_funcs.go
rahulbabu95: Makes sense. Yeah failing early is definitely preferred

tink/controller/internal/workflow/template_funcs.go
rahulbabu95: I am not too keen to the defaulting, since that's additional work on the consumer.

jacobweinstock (approved):
Thanks, @rahulbabu95

PR #418, opened 2025-10-22
Fix comment example for formatPartition function

magicaljellybeans:
Corrected example in comments for formatPartition function.


jacobweinstock (approved):
Thanks, @magicaljellybeans

PR #493, opened 2025-12-17
hardware.go: avoid using `{{` in description for References field in HardwareSpec

rpardini:
- that ends up in the CRD yaml
- if that is used in a Helm Chart, eg, a future `tinkerbell-crds` chart...
- ... it causes error `executing "..../tinkerbell.org_hardware.yaml" at <.references.lvm>`
- just avoid it for now
Signed-off-by: Ricardo Pardini <ricardo@pardini.net>


rpardini:
"very" minor.

PR #509, opened 2026-01-04
feat(template): add secretRef to templates

parsa97:
## Description

This PR adds secretRef to templates. This allows secrets to be referenced and rendered inside templates.
 
## How Has This Been Tested?

1. Deployed Tinkerbell in a k3d cluster.
2. Created a template that references a Secret in the tinkerbell namespace.
3. Created a workflow from this template.
4. Verified via kubectl describe workflow that secret values were correctly rendered from secretRef into the workflow.

None

## Checklist:

I have:

- [x] added unit or e2e tests


parsa97:
Hi @rpardini @jacobweinstock , Could we have a conclusion in this PR?

parsa97:
Hi @jacobweinstock , Thanks for collaborating in here, I will try to implement the second option; add References, as they are in the Hardware spec, to the Template spec. Then we can use the same access policies to secure them.

PR #532, opened 2026-01-25
fix: preserve hostname in --ipxe-script-tink-server-addr-port

### LINKED ISSUE #531: --ipxe-script-tink-server-addr-port hostname is ignored, always replaced with --public-ipv4
## Description

When setting `--ipxe-script-tink-server-addr-port` to a hostname (e.g., `reboot.example.com:443`), the hostname is discarded and replaced with the value of `--public-ipv4`. Only the port is preserved from the user-provided value.

This prevents using a hostname for the Tink server gRPC authority in iPXE scripts, which is required for:
- Split-horizon DNS deployments
- TLS certificate validation (certificates are issued for hostnames, not IPs)
- Routing through reverse proxies (e.g., Traefik) where the hostname determines routing

## Expected Behavior

Setting `--ipxe-script-tink-server-addr-port=reboot.example.com:443` should result in the iPXE script containing:

```ipxe
set grpc_authority reboot.example.com:443
```

## Actual Behavior

The iPXE script contains:

```ipxe
set grpc_authority 192.168.1.100:443
```

Where `192.168.1.100` is the value of `--public-ipv4`, regardless of what hostname was specified in `--ipxe-script-tink-server-addr-port`.

## Root Cause

In `cmd/tinkerbell/flag/smee.go`, the `Convert()` function (lines 205-211) extracts only the port and always uses `publicIP` for the host:

```go
s.Config.TinkServer.AddrPort = func() string {
    _, port := splitHostPort(s.Config.TinkServer.AddrPort)  // Host is discarded!
    if port == "" {
        port = fmt.Sprintf("%d", smee.DefaultTinkServerPort)
    }
    return fmt.Sprintf("%s:%s", publicIP.String(), port)  // Always uses publicIP
}()
```

## Suggested Fix

Only use `publicIP` as a fallback when no host is provided:

```go
s.Config.TinkServer.AddrPort = func() string {
    host, port := splitHostPort(s.Config.TinkServer.AddrPort)
    if port == "" {
        port = fmt.Sprintf("%d", smee.DefaultTinkServerPort)
    }
    // Only use publicIP as fallback when host is empty
    if host == "" && publicIP.IsValid() && !publicIP.IsUnspecified() {
        host = publicIP.String()
    }
    return fmt.Sprintf("%s:%s", host, port)
}()
```

## Reproduction Steps

1. Deploy Tinkerbell with Helm or run the binary directly
2. Set environment variables:
   ```bash
   export TINKERBELL_PUBLIC_IPV4=192.168.1.100
   export TINKERBELL_IPXE_SCRIPT_TINK_SERVER_ADDR_PORT=reboot.example.com:443
   export TINKERBELL_IPXE_SCRIPT_TINK_SERVER_USE_TLS=true
   ```
3. Boot a machine via PXE and examine the iPXE script served
4. Observe that `grpc_authority` is `192.168.1.100:443` instead of `reboot.example.com:443`

## Environment

- Tinkerbell version: v0.22.x (also affects main branch)
- Deployment: Helm chart on k3s
- File: `cmd/tinkerbell/flag/smee.go`

## Impact

- **Split-horizon DNS broken**: Users who need different IPs for LAN vs WAN clients cannot use a hostname to let DNS handle the resolution.

- **Reverse proxy routing broken**: When using Traefik/nginx to route gRPC traffic based on hostname, the connection goes to the wrong backend or fails entirely.

- **TLS certificate validation fails**: Certificates are typically issued for hostnames, not IPs. When HookOS/tink-worker connects to `192.168.1.100:443`, TLS validation fails because the cert is for `reboot.example.com`.

## Workarounds

Current workarounds (all suboptimal):

1. **Disable TLS verification** (`--ipxe-script-tink-server-insecure-tls=true`) - security risk
2. **Issue cert for IP address** - operationally complex, doesn't help split-horizon
3. **Serve custom static iPXE script** - bypasses Smee's dynamic script generation
4. **Traefik middleware to rewrite** - adds complexity

## Related

This behavior differs from other hostname settings like `--dhcp-ipxe-http-script-host` which correctly preserves the user-provided hostname.

---

## Checklist for PR

- [ ] Fix `Convert()` in `cmd/tinkerbell/flag/smee.go`
- [ ] Add unit test for hostname preservation
- [ ] Update documentation to clarify hostname support
- [ ] Test with both IP and hostname values
- [ ] Test with empty value (should fall back to publicIP)

hh:
## Summary

The `Convert()` function in `cmd/tinkerbell/flag/smee.go` was discarding the hostname from `--ipxe-script-tink-server-addr-port` and always replacing it with the `publicIP` value. This prevented using hostnames for the Tink server gRPC authority in iPXE scripts.

## Use Cases Fixed

- **Split-horizon DNS** - Different networks resolve the same hostname to different IPs
- **TLS certificate validation** - Certs are issued for hostnames, not IPs
- **Reverse proxy routing** - Hostname determines routing (e.g., Traefik IngressRoute)

## Changes

- Preserve user-provided hostname; only use `publicIP` as fallback when host is empty
- Added unit tests covering hostname preservation, port defaults, and fallback behavior

## Test Plan

- [x] `make ci` passes (lint, test, coverage)
- [x] Unit tests cover: hostname+port, hostname only, IP+port, empty input, port-only input

Fixes #531

### DIFF
--- a/cmd/tinkerbell/flag/smee.go
+++ b/cmd/tinkerbell/flag/smee.go
@@ -203,11 +203,15 @@ func (s *SmeeConfig) Convert(trustedProxies *[]netip.Prefix, publicIP netip.Addr
 	}
 
 	s.Config.TinkServer.AddrPort = func() string {
-		_, port := splitHostPort(s.Config.TinkServer.AddrPort)
+		host, port := splitHostPort(s.Config.TinkServer.AddrPort)
 		if port == "" {
 			port = fmt.Sprintf("%d", smee.DefaultTinkServerPort)
 		}
-		return fmt.Sprintf("%s:%s", publicIP.String(), port)
+		// Only use publicIP as fallback when host is empty/unset
+		if host == "" && publicIP.IsValid() && !publicIP.IsUnspecified() {
+			host = publicIP.String()
+		}
+		return fmt.Sprintf("%s:%s", host, port)
 	}()
 
 	// Set bind addresses if bindAddr is specified.
--- a/cmd/tinkerbell/flag/smee.go
+++ b/cmd/tinkerbell/flag/smee.go
@@ -109,6 +109,7 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 	fs.Register(IPXEHTTPScriptRetries, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.Retries, sc.Config.IPXE.HTTPScriptServer.Retries))
 	fs.Register(IPXEHTTPScriptRetryDelay, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.RetryDelay, sc.Config.IPXE.HTTPScriptServer.RetryDelay))
 	fs.Register(IPXEHTTPScriptOSIEURL, &url.URL{URL: sc.Config.IPXE.HTTPScriptServer.OSIEURL})
+	fs.Register(IPXEScriptSyslogFQDN, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.SyslogFQDN, sc.Config.IPXE.HTTPScriptServer.SyslogFQDN))
 	fs.Register(IPXEBinaryInjectMacAddrFormat, &ffval.Enum[constant.MACFormat]{
 		ParseFunc: macAddrFormatParser,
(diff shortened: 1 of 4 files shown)


jacobweinstock:
Resolved by https://github.com/tinkerbell/tinkerbell/pull/897. @hh , thanks for bringing this up!

PR #534, opened 2026-01-25
fix(smee): add syslog FQDN support for iPXE scripts

### LINKED ISSUE #533: add --ipxe-script-syslog-fqdn flag for iPXE scripts
### Problem

The `syslog_host` kernel parameter in iPXE scripts is always set from `--dhcp-syslog-ip`, which must be an IP address. This breaks global bare metal provisioning scenarios where:

1. Machines boot from different networks/locations worldwide
2. Split-horizon DNS is used
3. A resolvable hostname is needed instead of a LAN IP

Currently in `smee/smee.go`:
```go
jh := script.Handler{
    // ...
    PublicSyslogFQDN: c.DHCP.SyslogIP.String(),  // Always IP, never hostname
    // ...
}
```

### Use Case

Provisioning bare metal servers globally via Tinkerbell. The iPXE script needs:
C- `syslog_host=reboot.example.com` (**this issue**)

Without a hostname option, remote machines can't reach the syslog server because:
- The LAN IP (e.g., `192.168.1.100`) isn't routable globally
- DNS resolves `reboot.example.com` to the correct IP for each location

### Proposed Solution

Add a new flag `--ipxe-script-syslog-fqdn` that:
1. Sets the syslog hostname in iPXE scripts when provided
2. Falls back to `--dhcp-syslog-ip` when not set (backward compatible)

### Related

- #531 / #532 - Similar fix for `--ipxe-script-tink-server-addr-port`

hh:
## Description

Add `--ipxe-script-syslog-fqdn` flag to allow specifying a hostname for the syslog server in iPXE scripts, instead of always using the IP address from `--dhcp-syslog-ip`.

This is needed for global bare metal provisioning where machines in different locations need to reach the syslog server via a resolvable hostname rather than a LAN IP address.

The new field falls back to the existing SyslogIP when not set, maintaining backward compatibility.

Fixes: #533

## How Has This Been Tested?

- Unit tests added in `smee/smee_test.go` covering:
  - FQDN set overrides IP
  - Empty FQDN falls back to IP
  - FQDN with subdomain preserved
- Deployed to production cluster, verified iPXE script output shows `syslog_host=reboot.mcclimans.net` instead of IP

No impact - fully backward compatible. When `--ipxe-script-syslog-fqdn` is not set, behavior is identical to before (uses `--dhcp-syslog-ip`).

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/cmd/tinkerbell/flag/smee.go
+++ b/cmd/tinkerbell/flag/smee.go
@@ -109,6 +109,7 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 	fs.Register(IPXEHTTPScriptRetries, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.Retries, sc.Config.IPXE.HTTPScriptServer.Retries))
 	fs.Register(IPXEHTTPScriptRetryDelay, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.RetryDelay, sc.Config.IPXE.HTTPScriptServer.RetryDelay))
 	fs.Register(IPXEHTTPScriptOSIEURL, &url.URL{URL: sc.Config.IPXE.HTTPScriptServer.OSIEURL})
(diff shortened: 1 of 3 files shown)


jacobweinstock:
Resolved by https://github.com/tinkerbell/tinkerbell/pull/897. @hh , thanks for bringing this up!

PR #540, opened 2026-02-03
Allow templating in `customboot`

s3rj1k:
PR Allows templating MAC in `customboot` steps.

Available template variables:
  - `{{ (index .Hardware.Interfaces 0).DHCP.MAC }}`: First interface MAC address in dash-separated format (e.g., "52-54-00-12-34-01")
  - `{{ (index .Hardware.Interfaces 1).DHCP.MAC }}`: Second interface MAC address, etc.

Example workflow:

```
apiVersion: tinkerbell.org/v1alpha1
kind: Workflow
metadata:
  name: vm1
  namespace: tinkerbell-system
spec:
  bootOptions:
    bootMode: customboot
    custombootConfig:
      postActions:
      - virtualMediaAction:
          kind: CD
          mediaURL: ""
      - powerAction: "off"
      - bootDevice:
          device: disk
          efiBoot: true
          persistent: true
      - powerAction: "on"
      preparingActions:
      - powerAction: "off"
      - virtualMediaAction:
          kind: CD
          mediaURL: ""
      - virtualMediaAction:
          kind: CD
          mediaURL: http://172.17.1.1:7171/iso/{{ (index .Hardware.Interfaces 0).DHCP.MAC | replace ":" "-" }}/hook.iso
      - bootDevice:
          device: cdrom
          efiBoot: true
      - powerAction: "on"
    toggleAllowNetboot: true
  hardwareMap:
    device_1: "52:54:00:12:34:01"
  hardwareRef: vm1
  templateRef: vm1
```

### DIFF
--- a/docs/technical/BOOT_MODES.md
+++ b/docs/technical/BOOT_MODES.md
@@ -131,5 +131,110 @@ spec:
       - powerAction: "reset"
 ```
 
-[PREPARING Reference](/tink/controller/internal/workflow/pre.go#L158)  
+[PREPARING Reference](/tink/controller/internal/workflow/pre.go#L158)
 [POST Reference](/tink/controller/internal/workflow/post.go#L61)
+
+#### Example: PXE Boot with customboot
+
+The `customboot` mode can replicate the `netboot` behavior while allowing additional customization. This is useful when you need fine-grained control over the boot sequence:
+
+```yaml
+apiVersion: "tinkerbell.org/v1alpha1"
+kind: Workflow
+metadata:
+  name: example-pxe-boot
+spec:
+  templateRef: example
+  hardwareRef: example
+  bootOptions:
+    bootMode: customboot
+    custombootConfig:
+      preparingActions:
+      - powerAction: "off"
+      - bootDevice:
+          device: "pxe"
+          efiBoot: true  # Set based on your hardware requirements
+      - powerAction: "on"
+      postActions:
+      - powerAction: "off"
+      - bootDevice:
+          device: "disk"
+          persistent: true
+          efiBoot: true
+      - powerAction: "on"
+```
+
+This configuration will:
+1. Power off the machine
+2. Set the next boot device to PXE
+3. Power on the machine (boots from network)
+4. After workflow completion, power off the machine
+5. Set boot device back to disk persistently
+6. Power on the machine to boot from disk
+
+### Templating in customboot
+
+The `customboot` mode supports Go template syntax in action fields, enabling dynamic configuration based on Hardware specifications. This is particularly useful for virtual media URLs that need to include the Machine's MAC address.
+
+#### Available Template Data
+
+Templates have access to the full Hardware specification through the `.Hardware` variable:
+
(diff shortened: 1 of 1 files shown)


tink/controller/internal/workflow/pre.go
```diff
@@ -171,3 +198,67 @@ func (s *state) prepareWorkflow(ctx context.Context) (reconcile.Result, error) {
...
+		}
+	}
+
+	// Process each action
+	result := make([]bmc.Action, len(actions))
+	for i, action := range actions {
+		result[i] = action
+
+		// Template VirtualMediaAction.MediaURL if present
+		if action.VirtualMediaAction != nil && strings.Contains(action.VirtualMediaAction.MediaURL, "{{") {
```
jacobweinstock: Why is this needed?

```Go
strings.Contains(action.VirtualMediaAction.MediaURL, "{{")
```
`templateString()` can handle a string without any template directives.

tink/controller/internal/workflow/pre.go
```diff
@@ -171,3 +198,67 @@ func (s *state) prepareWorkflow(ctx context.Context) (reconcile.Result, error) {
 
 	return reconcile.Result{}, nil
 }
+
+// templateData holds data available for templating in customboot actions.
+type templateData struct {
+	// MACAddresses contains all interface MAC addresses in dash-separated format, indexed by interface order.
+	// Example: {{ index .MACAddresses 0 }} returns "52-54-00-12-34-01"
```
jacobweinstock: I'm concerned about this format (`{{ index .MACAddresses 0 }}`). We already have some [known template arguments/formats](https://tinkerbell.org/docs/v0.22/concepts/templates/#available-template-arguments). Introducing a new one feels disjointed to that and a less than ideal user experience. Could the same format defined in the docs be used?

While using `{{ (index .Hardware.Interfaces 0).DHCP.MAC }}` does feel a bit misleading because we would only want to allow access to the mac address even though a user could try to access other `.Hardware` locations, I still feel better about using the existing format.

Thoughts?

tink/controller/internal/workflow/pre.go
s3rj1k: True, I donno why I did that, lol, I've rebased and remove this line, resolving

tink/controller/internal/workflow/pre.go
s3rj1k: It does make sense to use existing format, still to make this a bit more transparent to consumer I would suggest implementing this like in change below

https://github.com/bellfork/tinkerbell/commit/6c2d60f16a2f710581552f87b37642ca3720a21c

tldr: `http://172.17.1.1:7171/iso/{{ (index .Hardware.Interfaces 0).DHCP.MAC | replace ":" "-" }}/hook.iso`

s3rj1k: (also I guess we should squash commits before merge and update PR description)

s3rj1k:
> Hey @s3rj1k . Thanks for this. Looks good. Mind adding a doc for this in `docs/technical/`?

Sure, added in new commit, if all looks good, we can probably squash everything into single commit during merge.

PR #549, opened 2026-02-09
build: migrate build system to GoReleaser

appkins:
## Description

Migrate the build pipeline to [GoReleaser](https://goreleaser.com/) for unified, reproducible binary and container image builds across all platforms.

**This PR has been scoped down** based on reviewer feedback. The original changes have been split into three focused PRs:

| PR | Scope |
|----|-------|
| **This PR (#549)** | GoReleaser build system (`.goreleaser.yaml`, CI workflow, Makefile build targets, Dockerfile paths) |
| [#682](https://github.com/tinkerbell/tinkerbell/pull/682) | Move tooling to `go.mod` tool directives; unify formatting with `golangci-lint` |
| [#683](https://github.com/tinkerbell/tinkerbell/pull/683) | Static/distroless Docker images with compiled ipmitool |

---

### Changes in this PR

**`.goreleaser.yaml`** (new)
- Build `tinkerbell`, `tinkerbell-embedded` (build tag: `embedded`), and `tink-agent` for `linux/amd64` and `linux/arm64` with `CGO_ENABLED=0`.
- UPX compression (`best+lzma`) for all binaries.
- Per-artifact checksums.
- Docker images built from `Dockerfile.tinkerbell` / `Dockerfile.agent` with OCI labels.
- Branch-aware tagging: `v{{ .Version }}` on tags, `sha-{{ .ShortCommit }}` on main, `edge` for nightly.
- Cosign image signing (`docker_signs`).
- `git.ignore_tags` filters `api/v*` and `legacy-tink-v*` tags.
- `release.disable: true` — GoReleaser handles the full release when `IS_RELEASE=true`.

**`Makefile`**
- Add `GORELEASER_VER`, `GORELEASER` binary path (`bin/goreleaser-<ver>`).
- Add `GITHUB_REPOSITORY_OWNER` variable so forks publish charts to their own registry.
- Replace `build: out/tinkerbell` with `goreleaser build --auto-snapshot`.
- Add `cross-compile` alias for backward compatibility.
- Add `build-image` and `build-image-push` targets using GoReleaser.
- **Preserve** `image`, `image-agent`, `prepare-buildx`, `build-push-image*` targets for local Docker builds without a registry push (addresses reviewer concern about losing local image builds).
- Add `$(GORELEASER)` install recipe alongside existing tools.
- Update `clean` to also remove `dist/`.
- Remove `GIT_TAG` block — GoReleaser derives the version from git tags.

**`.github/workflows/ci.yaml`**
- **Fixes reviewer concern**: Build job now runs on **all PRs**, not just `main`/tags.
  - On PRs: `goreleaser build` (compiles binaries, no Docker push)
  - On `main`: `goreleaser release --skip=validate` (builds + pushes images)
  - On tags: `goreleaser release` (full release with changelog + signing)
- Docker login, QEMU, Buildx, Cosign only installed when actually pushing.
- Remove `build-binaries` matrix job, manual artifact upload/download, and `release-notes` job.
- Add `generate` job to verify code generation is up-to-date on every PR.
- Upload GoReleaser `dist/` artifacts for inspection.

**`Dockerfile.agent`**
- Update `COPY` path from `out/tink-agent-${TARGETOS}-${TARGETARCH}` to `${TARGETOS}/${TARGETARCH}/tink-agent` to match GoReleaser `dist/` layout.

**`.github/mergify.yml`**
- Update check-success conditions to reference new CI job names (`build-publish-container-images`, `generate`).

**`.gitignore`**
- Add `bin/`, `dist/`, `junit.xml`.

---

### Addressing reviewer comments

| Comment | Response |
|---------|----------|
| "Build job only runs on main and tags — need early error detection on PRs" | Fixed: `build-publish-container-images` now runs on all PRs. On PRs it runs `goreleaser build` (binaries only), so compilation is validated before merging. |
| "We are losing local image builds" | Fixed: `make image` and `make image-agent` (direct `docker build`) are preserved alongside the new `make build-image` (GoReleaser). |
| "Not using `make` targets in CI" | Fixed: `generate` job uses `make generate`. |
| "`-X main.version={{.Version}} -X main.commit=...`" | These ldflags are required by GoReleaser for SBOM generation and Cosign signing metadata. They complement `pkg/build` — GoReleaser injects the actual version at release time. |
| "`api/go.mod` must be its own module" | Reverted: `api/go.mod` is untouched in this PR. |

---

## How Has This Been Tested?

- `goreleaser build --clean --auto-snapshot` builds binaries for both `amd64` and `arm64`.
- `goreleaser release --clean --skip=validate --skip=sign --auto-snapshot` builds Docker images locally.
- `make build` and `make cross-compile` work as expected.

## Checklist

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,9 +11,59 @@ env:
   REGISTRY: ghcr.io
   IMAGE: ghcr.io/${{ github.repository }}
   CGO_ENABLED: 0
-  GO_VERSION: "1.26"
+  GITHUB_REPOSITORY_OWNER: ${{ github.repository_owner }}
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v9
+        with:
+          args: --timeout=10m
+          version: v2.11.2
+
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v9
+        with:
+          install-only: true
+          version: v2.11.2
+
+      - name: Run go mod tidy
+        run: go mod tidy
+
+      - name: Generate code and format
+        run: make generate
+
+      - name: Check for changes
+        run: git diff --compact-summary --exit-code
+
+      - name: Check for new files
+        run: |
+          if git status --porcelain | grep -qE '^(\?\?|A )'; then
+            echo "Error: new files present (untracked or staged)" >&2
+            exit 1
+          fi
+
   validation:
     runs-on: ubuntu-latest
     steps:
@@ -26,71 +76,44 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v6
         with:
-          go-version: "${{ env.GO_VERSION }}"
-          cache: false
+          go-version-file: 'go.mod'
 
-      - name: Restore Go cache
-        uses: actions/cache/restore@v5
+      - name: Run Tests
+        if: always()
+        run: go tool gotestsum --junitfile junit.xml -- -coverprofile=coverage.txt -covermode=atomic -race -count=1 -v ./...
+        env:
+          CGO_ENABLED: 1
+
+      - name: Generate HTML coverage report
+        if: always()
+        run: go tool cover -html=coverage.txt -o coverage.html
+
+      - name: Upload coverage report
+        if: always()
+        uses: actions/upload-artifact@v4
         with:
+          name: coverage-report
           path: |
-            ~/go/pkg/mod
-            ~/.cache/go-build
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-            ${{ runner.os }}-go-
+            coverage.txt
+            coverage.html
+            junit.xml
 
-      - name: Fix no space errors and Fetch Deps
-        run: |
-          # fixes "write /run/user/<uid>/<nonce>: no space left on device" error
-          sudo mount -o remount,size=3G "/run/user/$(id -u)" || true
-          go mod download
-
-      - name: Generate UI assets
-        run: make ui-generate
-
-      - name: Run all CI checks, linting, tests, etc
-        run: make ci TEST_ARGS="-count=1"
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/test-results-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
 
       - name: upload to codecov
-        uses: codecov/codecov-action@v6
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
-      - name: Trim Go cache
-        if: ${{ github.ref == 'refs/heads/main' }}
-        shell: bash
-        # As the go command works, it either creates build cache files or touches
-        # ones it uses at most once an hour. When it trims the cache, it trims
-        # files that have not been modified/touched in 5+ days.
-        # To keep our saved cache lean, trim all files except ones that were just
-        # created/touched as part of this run.
-        run: |
-          find ~/.cache/go-build -type f -mmin +90 -delete
-
-      - name: Set Go cache date
-        shell: bash
-        run: echo "GO_CACHE_DATE=$(date +%Y-%m-%d)" >> $GITHUB_ENV
-
-      - name: Save Go cache
-        if: ${{ github.ref == 'refs/heads/main' }}
-        uses: actions/cache/save@v5
+        uses: codecov/codecov-action@v5
         with:
-          # Caches both the downloaded modules and the compiled build cache.
-          path: |
-            ~/go/pkg/mod
-            ~/.cache/go-build
-          # Save to eg Linux-go-$hash-YYYY-MM-DD to keep the cache fresh
-          key: "${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ env.GO_CACHE_DATE }}"
+          token: ${{ secrets.CODECOV_TOKEN }}
 
   validate-helm-chart:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
-          fetch-tags: true
+        uses: actions/checkout@v5
 
       - name: install helm
         uses: Azure/setup-helm@v5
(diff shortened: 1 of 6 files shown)


appkins:
I have a bunch of well tested code to push upstream and figured I would start with this one since it won't be changing on my fork. 

- Cleans up the file tree to work properly with automated linting/formatting.
- Simplifies the build process with goreleaser - single config to manage three binary releases
- Creates distroless versions of the images that are ~4Mb each
- Supports building Tinkerbell in forks for easier testing

.github/workflows/ci.yaml
```diff
@@ -121,65 +62,59 @@ jobs:
...
-          path: |
-            out/${{ matrix.name }}-*
-            ${{ matrix.extra-paths }}
-          if-no-files-found: error
+          version: v3.17.0
+
+      - name: Lint and Template Helm chart
+        run: make helm-lint helm-template
 
   build-publish-container-images:
```
jacobweinstock, line 126: This is the only job that builds the binaries and container images and this job only runs on main and tags. This means that we don't know that the build process actually works and the code compiles until it hits main. This must be done on all PRs. We need to catch these errors early.

.goreleaser.yaml
```diff
@@ -0,0 +1,130 @@
...
+    main: ./cmd/tinkerbell
+    binary: tinkerbell
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    no_unique_dist_dir: false
+    ldflags:
+      - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser
```
jacobweinstock, line 24: This is new. Version and commit are currently captured through the `pkg/build` package. What does this provide?

`-X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser`

.goreleaser.yaml
```diff
@@ -0,0 +1,130 @@
...
+      - -extldflags '-static'
```
jacobweinstock: This shouldn't be needed. Setting `CGO_ENABLED=0` is enough to produce a statically linked binary.

Reference: https://groups.google.com/g/golang-nuts/c/GGYpabubcnk

Makefile
```diff
@@ -90,9 +106,8 @@ vet: ## Run go vet
 	go vet ./...
 
 .PHONY: fmt
-fmt: $(GOIMPORTS_FQP) ## Run go fmt
-	go fmt ./...
-	$(GOIMPORTS_FQP) -w .
+fmt: $(GOLANGCI_LINT) ## Run go fmt
+	$(GOLANGCI_LINT) fmt ./...
```
jacobweinstock: what is gained by using golangci-lint for the formatting over `go fmt` and `goimports`?

Makefile
```diff
@@ -35,43 +43,51 @@ ifeq ($(LOCAL_ARCH),x86_64)
...
-GODEPGRAPH_FQP := $(TOOLS_DIR)/godepgraph-$(GODEPGRAPH_VER)
+CONTROLLER_GEN_VER	:= v0.20.0
+BUF_VER							:= v1.56.0
+UPX_VER							:= 4.2.4
+GODEPGRAPH_VER			:= v0.0.0-20240411160502-0f324ca7e282
+GOLANGCI_LINT_VER		:= v2.8.0
+GORELEASER_VER			:= v2.12.2
+
+# Directories.
+TOOLS_BIN_DIR := $(abspath bin)
```
jacobweinstock, line 56: you've moved away from `out/tools`. what's the motivation for this?

Makefile
```diff
@@ -209,78 +178,76 @@ helm-template: ## Helm template for Tinkerbell
...
-prepare-buildx: ## Prepare the buildx environment.
-## the "|| true" is to avoid failing if the builder already exists.
-	docker buildx create --name tinkerbell-multiarch --use --driver docker-container || true
-
-.PHONY: image
-image: cross-compile ## Build the Tinkerbell container image
-	docker build -t $(IMAGE_NAME) -f Dockerfile.tinkerbell .
+.PHONY: build-image
+build-image: | $(GORELEASER) ## Build the container images
+	$(GORELEASER) release --clean --skip=sign --verbose
```
jacobweinstock: I tried this and it looks like it requires the images to be pushed to a registry. We are losing local image builds. I prefer we don't lose local image builds.

.github/workflows/ci.yaml
```diff
@@ -11,106 +11,47 @@ env:
...
         with:
-          version: v3.17.0
+          lint: ${{ github.event_name == 'push' }}
+          format: ${{ github.event_name == 'push' }}
+          breaking: ${{ github.event_name == 'push' }}
 
-      - name: Lint and Template Helm chart
-        run: make helm-lint helm-template
+      - name: Generate Protobuf code
+        run: buf generate
```
jacobweinstock: Not using the `make` targets we have, `make generate-proto` in this instance, means that CI and local development aren't guaranteed to be running the same thing and requires us to keep to locations in sync. I've found this to be difficult for reproducing issues found in CI and adds a maintenance burden to keep these 2 locations in sync.

go.mod
```diff
@@ -314,3 +314,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.34.3
 	k8s.io/sample-controller => k8s.io/sample-controller v0.34.3
 )
+
+replace github.com/tinkerbell/tinkerbell => ./
```
jacobweinstock: What's the motivation for this change?

appkins: This allows local builds and CI/CD to work in forked environments.

Makefile
appkins: These are used as vendored binaries for automation processes, not for distribution. I would prefer a distribution directory that is designed to be packaged for release with just the necessary ingredients.

Meanwhile, `bin` is a POSIX standard for pre-built binaries.

Makefile
appkins: It's a unified tool that streamlines linting to prevent inconsistencies. It helps avoid significant refactoring when submitting PRs and having the full linting run for the first time. In conjunction with `golangci-lint run --fix`, this should represent  a single tool for all `*.go` file standardization.

.goreleaser.yaml
appkins: I believe this is tied into goreleaser's process for cosigning and sbom generation. In my forked repository, all packages are signed with a bill of materials (when run through automation). I have been running goreleaser with a handful of lifecycle skips lately however.

go.mod
appkins: It does nothing I removed it. Good catch

appkins:
@jacobweinstock I submitted a full re-write of the CI/CD system that unifies the dependency management.

Something I noticed when building and rebasing was that often times tooling versions are different from the library version included in the go modules. 

The latest changes fix all the previous issues with goreleaser including the prefix in the tags like `api/`.

Additionally, the generation commands are completely streamlined and can be automated with `go generate ./...`.

The CI workflow adds checks fr PRs like `golangci-lint` that can comment on issues if you decide to set up permissions for the workflow to comment in the repo.

The golangci-lint config was tricky to iron out but now respects the generated files while still performing functions like import sorting.

api/go.mod
jacobweinstock, line 1: this must be its own module. https://github.com/tinkerbell/tinkerbell/pull/186

appkins:
> Would you willing to break these changes into the smallest units of change possible?

@jacobweinstock 
How does this sound?

1. PR for `goreleaser` as the build system
2. PR for moving the tooling to `go.mod` as tools and unifying the formatting/linting with `golangci-lint`
3. Static, distress, minimal Docker images with static IPMI binary 

This was my attempt at breaking my code into the smallest unit of change lol.

I have a huge backlog of features I would like to submit but face hell every time I have to rebase due to version changes upstream. My hope is that this will streamline the third party tooling so it can be predictably maintained locally - leaving fewer conflicts to resolve.

---

> There's moving to goreleaser, moving all checks to golangci-lint, refactoring how tools are defined and downloaded, and statically building ipmitool

Goreleaser,  golangci-lint and the tooling refactor are all related and I would prefer to not break these up. The static IPMI image can be easily split out though.

Goreleaser *can* run independently but is one of two total binaries that must be present on the host machine for the build to succeed. The idea is that all other dependencies are transient, while the build system and linter should be present.

On the other hand, `golangci-lint` and the tooling refactor should really stay together. This piece is very intentional. Removing the requirement to have many binaries installed on the host machine is primarily possible thanks to `golangci-lint` bundling every formatting and linting function into one binary. The other tools are required for code generation and included in the go.mod as tools.

---

> And, locally running `make build` doesnt work for me. Multiple issues.

> I had to manually run `make tools`, but it still didnt work.
> 

I'll take a look at this. Golangci-lint recommends not installing it as a tool since there are so many parts. I must have had it cached from before and swapped it over to the vendor pattern. Besides golanglint-ci, the other components are very stable due to Go's ecosystem.

appkins:
@jacobweinstock 
> I had to manually run make tools, but it still didnt work.

Can you try the build with the new changes? It was just missing a Make prerequisite which I added

api/go.mod
appkins: @jacobweinstock I just tested switching the dependency in `cluster-api-provider-tinkerbell` and it adds no new dependencies - just updates them to the versions in the root level go.mod

appkins: In other words, `github.com/tinkerbell/tinkerbell/api` changes to `github.com/tinkerbell/tinkerbell` in capi-provider's go.mod. After running `go mod tidy`, the listed dependencies are identical, with new versions to match the upstream project.

jacobweinstock: There might not be any conflicts right now, but this couples all dependency versions. When Tinkerbell moves controller-runtime (a particularly difficult dependency) version but CAPT is not ready for that version there is no way for CAPT to get the updated API definitions. This really must be its own module.

PR #655, opened 2026-03-19
feat: add per-action Linux namespaces (network and PID) support

juninhoojl:
## Description

Adds a per-action `namespaces` object (mirroring the unpublished v1alpha2 API spec) that lets a workflow action choose the Linux **network** and **PID** namespaces its container runs in. Setting `namespaces.network: "host"` runs the action in the host's network namespace — needed by inventory tools like `lshw` that can only enumerate all host interfaces when sharing the host network stack.

The namespace values are passed through to the container runtime **as-is** — the server imposes no semantics on them, so any value the runtime accepts works (not just `"host"`).

The field flows through the full stack: workflow YAML parsing → controller → Kubernetes API status → protobuf/gRPC transport → Docker and containerd runtimes.

## Changes

- **`tink/controller/internal/workflow/types.go`** — Add `Namespaces` (network/pid) to the workflow YAML Action struct, documenting the pass-through behavior.
- **`api/v1alpha1/tinkerbell/workflow.go`** — Add `Namespaces *ActionNamespace` (a pointer, so `namespaces` is omitted from status when unset) and the `ActionNamespace` type. Deprecate the legacy top-level `Pid` field with a kubebuilder deprecation warning pointing to `namespaces.pid`.
- **`tink/controller/internal/workflow/convert.go`** — Propagate `namespaces` from parsed YAML into the Workflow status, only when at least one value is set.
- **`pkg/proto/get_action_response.proto`** — Add a `Namespaces` message (`network`, `pid`) and a `namespaces` field on `ActionResponse`. The top-level `pid` field is retained for backward compatibility and documented as deprecated.
- **`tink/server/internal/grpc/grpc.go`** — Populate `ActionResponse.Namespaces`, passing values through as-is; `namespaces.pid` takes precedence over the deprecated top-level `pid`.
- **`tink/agent/internal/transport/grpc/grpc.go`** — Read namespaces from the new object, falling back to the deprecated top-level `pid`.
- **`tink/agent/internal/runtime/docker/docker.go`** — Set `HostConfig.NetworkMode` from `namespaces.network` for any value (not only `"host"`).
- Regenerated protobuf, deepcopy, and the v1alpha1 CRD.

The **containerd** runtime already consumes `namespaces.network` for host networking, so no runtime change was required there.

## How Has This Been Tested?

- Unit tests across the controller (convert), server gRPC handler, and agent gRPC transport, covering an action with `namespaces.network: "host"` and `namespaces.pid` taking precedence over the legacy top-level `pid`.
- `go build ./...` and `go test ./...` pass (aside from environment-specific path tests that only pass on Linux).
- End-to-end: a Lima VM running Tinkerbell with a QEMU VM booting HookOS. A workflow with `namespaces.network: "host"` ran an inventory action that collected hardware inventory and posted it to an external API. A second workflow **without** `namespaces` verified backward compatibility.

### Example workflow YAML

```yaml
tasks:
  - name: "inventory"
    worker: "52:54:00:aa:bb:01"
    volumes:
      - /dev:/dev
      - /sys:/sys:ro
    actions:
      - name: "run-inventory"
        image: example/inventory:latest
        timeout: 300
        namespaces:
          network: "host"
        command:
          - inventory
        environment:
          ENDPOINT: "https://example.com/"
```

### How are existing users impacted? What migration steps/scripts do we need?

No impact. The `namespaces` field is optional and defaults to nil — existing workflows continue to use the runtime's default networking unless `namespaces.network` is explicitly set. The legacy top-level `pid` field still works; `namespaces.pid` is preferred when both are set. No migration steps or scripts are required.

## Checklist:

I have:

- [x] added unit or e2e tests


tink/agent/internal/runtime/docker/docker.go
```diff
@@ -78,6 +78,12 @@ func (c *Config) Execute(ctx context.Context, a spec.Action) error {
 	if a.Namespaces.PID != "" {
 		hostCfg.PidMode = container.PidMode(a.Namespaces.PID)
 	}
+	// When host network mode is requested, set NetworkMode to "host" so the
+	// container shares the host's network namespace. This is needed for tools
+	// like lshw that require full visibility of host network interfaces.
+	if a.Namespaces.Network == "host" {
+		hostCfg.NetworkMode = "host"
+	}
```
jacobweinstock: With this only host network is allowed through. What's the motivation for not allowing any network through? Maybe this can be `hostCfg.NetworkMode = a.Namespaces.Network`? thoughts?

tink/server/internal/grpc/grpc.go
```diff
@@ -321,6 +321,13 @@ func (h *Handler) doGetAction(ctx context.Context, req *proto.ActionRequest, opt
 		}(),
 		Pid: toPtr(action.Pid),
 	}
+	if action.Namespaces.Network != "" {
+		ar.HostNetwork = toPtr(action.Namespaces.Network == "host")
+	}
```
jacobweinstock: if i understand correctly, this make any network defined be used as the "host" network. This is not clear behavior from the spec and the spec's code comments. I'd prefer not doing this and letting any value be used. thoughts?

tink/server/internal/grpc/grpc.go
```diff
@@ -321,6 +321,13 @@ func (h *Handler) doGetAction(ctx context.Context, req *proto.ActionRequest, opt
...
+	// Prefer namespaces.pid over the top-level pid field.
+	if action.Namespaces.PID != "" {
+		ar.Pid = toPtr(action.Namespaces.PID)
+	}
```
jacobweinstock: I don't think the Spec code comments reflect this behavior. This needs to be surfaced to users.

jacobweinstock (commented):
Thanks, @juninhoojl. I left some comments. Also, should the containerd runtime be updated too?

api/v1alpha1/tinkerbell/workflow.go
```diff
@@ -248,6 +248,7 @@ type Action struct {
 	Command           []string          `json:"command,omitempty"`
 	Volumes           []string          `json:"volumes,omitempty"`
 	Pid               string            `json:"pid,omitempty"`
```
jacobweinstock, line 252: Mind adding this deprecation notice?

```Go
// Deprecated: This field is deprecated and will be removed in a future release. Use namespaces.pid instead.
// +kubebuilder:deprecatedversion:warning="Pid is deprecated and will be removed in a future release. Use namespaces.pid instead."
Pid               string            `json:"pid,omitempty"`
```

pkg/proto/get_action_response.proto
```diff
@@ -48,6 +48,11 @@ message ActionResponse {
     * Set the namespace that the process IDs will be in.
     */
    string pid = 11;
+   /*
+    * The network namespace for the action container.
+    * When set to "host", the container shares the host's network namespace.
+    */
+   string network = 12;
```
jacobweinstock: We should add a `namespaces` object with `pid` and `network` in it so it matches the API spec.

jacobweinstock:
Hey @juninhoojl , sorry for the delayed review here. This is really close. There's a couple of git merge conflicts that need resolved and a couple new comments to address and we can get this in. Thanks again for this!

juninhoojl:
Hello, @jacobweinstock! I’m working through all the points you requested today. Thanks for the reminder!

juninhoojl:
Hi @jacobweinstock! When you have a chance, I believe this addresses all the points you mentioned in your review.

I’m also planning to open a much smaller PR with a fix for the Smee syslog issue I mentioned on Tuesday, regarding logs being dropped.

Once again, thank you for taking the time to review this and for all your feedback!

pkg/proto/get_action_response.proto
juninhoojl: Done, added a `namespaces` object (`network` + `pid`) to match the API spec.

api/v1alpha1/tinkerbell/workflow.go
juninhoojl: Done, added the deprecation notice exactly as suggested.

api/v1alpha1/tinkerbell/workflow.go
```diff
@@ -256,6 +259,24 @@ type Action struct {
 	Message           string            `json:"message,omitempty"`
 }
 
+// ActionNamespace defines the Linux namespaces an action container runs in.
+// This mirrors the v1alpha2 API spec.
+type ActionNamespace struct {
```
jacobweinstock: `ActionNamespace` -> `ActionNamespaces`?

tink/controller/internal/workflow/types.go
```diff
@@ -29,4 +29,18 @@ type Action struct {
 	Volumes     []string          `yaml:"volumes,omitempty"`
 	Environment map[string]string `yaml:"environment,omitempty"`
 	Pid         string            `yaml:"pid,omitempty"`
+	Namespaces  ActionNamespace   `yaml:"namespaces,omitempty"`
+}
+
+// ActionNamespace defines the Linux namespaces an action container runs in.
+// This mirrors the v1alpha2 API spec.
+type ActionNamespace struct {
```
jacobweinstock: `ActionNamespace` -> `ActionNamespaces`?

jacobweinstock (changes requested):
Thanks, @juninhoojl . almost there. Just some naming inconsistencies to address. Thanks!

juninhoojl:
Thanks for the review! Done, I renamed it to ActionNamespaces.

jacobweinstock (approved):
Thanks, @juninhoojl !

PR #658, opened 2026-03-24
Only add release tag to existing OCI images, don't rebuild:

jacobweinstock:
## Description

Tag pushes were rebuilding all binaries and container images from scratch, duplicating work already done when the commit landed on main. Now tag pushes only build embedded binaries (for the GitHub release) and uses crane to add the release version and latest tags to the existing OCI images found by commit SHA.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock:
closing this because we lose the tag version inside the Go binary with this model. This is why we rebuild on tags. I prefer the existing model after working through this.

PR #723, opened 2026-04-25
Filter compatible BMC providers before open

### LINKED ISSUE #461: bmclib client to filter for compatible drivers
Raising this as an issue to point out, it would be ideal if this used the `FilterForCompatible` method
here https://github.com/tinkerbell/rufio/blob/13eef9a71caf1171e8f7aae2c85ffbb91379df9b/controllers/bmcclient.go#L19

This way the BMC is 'probed' for compatibility and only drivers known to be working with the BMC will be attempted
when the caller invokes the client methods. 

```
drivers := client.Registry.FilterForCompatible(ctx)
if len(drivers) == 0 {
  return nil, errors.New("no compatible bmclib drivers found")
}

client.Registry.Drivers = drivers
```

I understand the `redfish` driver is hardcoded at the moment, so whenever that is changed I suggest using the filter method above.

AkashKumar7902:
## Summary
- filter Rufio bmclib clients with `FilterForCompatible` before opening connections
- keep configured preferred provider ordering before filtering
- return a clear error when no compatible providers remain

## Testing
- `go test ./rufio/internal/controller -count=1`
- `go test ./rufio/... -count=1`
- `go vet ./rufio/...`
- `go test ./rufio/internal/controller -run TestPrepareClientForOpen -race -count=1`
- `TMPDIR=/home/akash/.cache make test TEST_PKG=rufio/internal/controller`

Resolves #461

### DIFF
--- a/rufio/internal/controller/client.go
+++ b/rufio/internal/controller/client.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"time"
@@ -14,6 +15,8 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/tinkerbell/tinkerbell/api/v1alpha1/bmc"
 )
+
+var errNoCompatibleDrivers = errors.New("no compatible BMC providers found")
 
 // ClientFunc defines a func that returns a bmclib.Client.
 type ClientFunc func(ctx context.Context, log logr.Logger, hostIP, username, password string, opts *BMCOptions) (*bmclib.Client, error)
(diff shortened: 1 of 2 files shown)


jacobweinstock:
Hey @AkashKumar7902, thanks for this! Unfortunately, issue #461 was lacking context. I've added more [here](https://github.com/tinkerbell/tinkerbell/issues/461#issuecomment-4661456365), but because of that missing context, we led you toward an implementation that isn't the direction we want to take. Apologies for that.

There's a new path forward that involves updating the `machine.bmc` CRD and the Rufio controller. I'm happy to work through those changes with you, but they aren't as straightforward as what you've done here. There's no expectation that you take this on, but if you decide to, please let me know how I can help.

Regardless, thanks for taking the time to contribute with this PR!

PR #765, opened 2026-05-20
Change whether the service is enabled

sepauli:
## Description

Certain deployments require hostNetwork alongside an enabled service.
This PR updates the behavior so that the service can be enabled or disabled without depending on hostNetwork

## How Has This Been Tested?

This change has been tested with an installation using service.enabled=true and service.enabled=false to verify whether the service is created.

Some users may need to disable the service explicitly.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/helm/tinkerbell/templates/service.yaml
+++ b/helm/tinkerbell/templates/service.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.deployment.hostNetwork }}
+{{- if .Values.service.enabled }}
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


jacobweinstock:
Hey @sepauli. Thanks for this. I fixed the unrelated lint failures. A rebase from main should get this all green.

jacobweinstock (approved):

jacobweinstock:
@mergifyio queue

PR #788, opened 2026-05-29
fix(ui): prevent duplicate select-all listeners after HTMX swap

anurag-p6:
## Summary
Every `htmx:afterSwap` calls `setupCheckAllFunctionality()`, which was adding new `change` listeners to `#selectAll` and `#selectAllContent` without removing the previous ones. After a few paginations or content swaps, clicking "select all" would fire the handler multiple times.
This PR hoists `handleSelectAll` and `updateSelectAllState` out of `setupCheckAllFunctionality` so the same function references can be passed to `removeEventListener`, and removes the old listener before re-attaching on each swap.
Also calls `updateSelectAllState()` at the end of `handleSelectAll` so both select-all checkboxes stay in sync when either one is clicked.
## Changes
- Move `handleSelectAll` and `updateSelectAllState` to stable outer scope
- Add `removeEventListener` before `addEventListener` on both select-all checkboxes
- Sync both select-all checkboxes after a select-all click

### DIFF
--- a/ui/assets/js/app.js
+++ b/ui/assets/js/app.js
@@ -472,41 +472,46 @@ document.addEventListener('DOMContentLoaded', () => {
 	updatePaginationURLs();
 
 	// Check all functionality
+	function updateSelectAllState() {
+		const selectAllMain = document.getElementById('selectAll');
+		const selectAllContent = document.getElementById('selectAllContent');
+		const checkedBoxes = document.querySelectorAll('.row-checkbox:checked');
+		const allRowCheckboxes = document.querySelectorAll('.row-checkbox');
+
+		if (allRowCheckboxes.length === 0) return;
+
+		const allChecked = checkedBoxes.length === allRowCheckboxes.length;
+		const someChecked = checkedBoxes.length > 0;
+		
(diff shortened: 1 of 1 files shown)


anurag-p6:
hi @jacobweinstock , is there any issue with this pr would? if yes would like to improve on that.

jacobweinstock:
> hi @jacobweinstock , is there any issue with this pr would? if yes would like to improve on that.

Hey @anurag-p6 . No, no issues. Please rebase from main and the failed linting will be resolved and then we can merge this. Thanks for this contribution!

anurag-p6:
Done with it @jacobweinstock, thank you.

jacobweinstock (approved):
Thanks, @anurag-p6 !

jacobweinstock:
Ah, @anurag-p6, looks like you need to to have GitHub verified signatures setup. 

https://docs.github.com/en/authentication/managing-commit-signature-verification

anurag-p6:
ops, thanks for assisting @jacobweinstock I've done the Github Signature Verification now.

jacobweinstock (approved):

PR #812, opened 2026-06-18
Adds first-class IPv6 and DHCPv6 support to Smee.

brk0v:
## Intro

I’d like to start a conversation about possible IPv6 support, following the direction described in roadmap issue [#44](https://github.com/tinkerbell/roadmap/issues/44).

I’m aware of the new `v1alpha2` CRDs and that this work is still in progress. In the meantime, I’d like to start the discussion with some code and gather early feedback, which could be reuse with `v1alpha2`.

Thank you for your time and attention.

---

## Summary

Adds first-class IPv6 and DHCPv6 support to Smee.

## IPv6 Infra Requirements and Limitations

The network must provide a stable way for Tinkerbell to identify client MAC addresses.
 - Direct clients need an EUI-64-derived IPv6 source address.
 - Relayed clients should include DHCPv6 Option 79, alternatively, the relay peer address must be EUI-64-derived.

## DHCPv6 Modes

- `stateless`: replies with DHCPv6 stateless configuration and netboot data for clients with matching Hardware records. Address assignment is handled outside Tinkerbell by RA/SLAAC.
- `auto-stateless`: extends stateless mode for unknown clients, allowing default netboot responses and static fallback iPXE scripts without pre-created Hardware records. This mode is the DHCPv6 analogue of DHCPv4 `auto-proxy`.
- `reservation`: assigns the IPv6 address reserved on the matching Hardware interface through DHCPv6 `IA_NA` flows.
- `derived`: derives a temporary boot-time IPv6 address from the client MAC address and configured direct or relay prefixes when no IPv6 reservation exists.

## Tests

Tested with playground locally and with real hardware servers under IPv6 only network.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/cmd/tinkerbell/flag/global.go
+++ b/cmd/tinkerbell/flag/global.go
@@ -17,6 +17,7 @@ type GlobalConfig struct {
 	OTELInsecure         bool
 	TrustedProxies       []netip.Prefix
 	PublicIP             netip.Addr
+	PublicIPv6           netip.Addr
 	BindAddr             netip.Addr
 	HTTPPort             int
 	HTTPSPort            int
@@ -71,6 +72,7 @@ func RegisterGlobal(fs *Set, gc *GlobalConfig) {
 	fs.Register(OTELEndpoint, ffval.NewValueDefault(&gc.OTELEndpoint, gc.OTELEndpoint))
 	fs.Register(OTELInsecure, ffval.NewValueDefault(&gc.OTELInsecure, gc.OTELInsecure))
 	fs.Register(PublicIP, &ntip.Addr{Addr: &gc.PublicIP})
+	fs.Register(PublicIPv6, &ntip.Addr{Addr: &gc.PublicIPv6})
 	fs.Register(TLSCertFile, ffval.NewValueDefault(&gc.TLS.CertFile, gc.TLS.CertFile))
 	fs.Register(TLSKeyFile, ffval.NewValueDefault(&gc.TLS.KeyFile, gc.TLS.KeyFile))
 	fs.Register(DisableHTTPToHTTPSRedirect, ffval.NewValueDefault(&gc.TLS.DisableHTTPToHTTPSRedirect, gc.TLS.DisableHTTPToHTTPSRedirect))
@@ -140,7 +142,12 @@ var TrustedProxies = Config{
 
 var PublicIP = Config{
 	Name:  "public-ipv4",
-	Usage: "public IPv4 address to use for all enabled services",
+	Usage: "public IPv4 address to advertise to clients",
+}
+
+var PublicIPv6 = Config{
+	Name:  "public-ipv6",
+	Usage: "public IPv6 address to advertise to clients",
 }
 
 var EnableSmee = Config{
@@ -195,7 +202,7 @@ var EnableCRDMigrations = Config{
 
 var BindAddr = Config{
 	Name:  "bind-address",
-	Usage: "IP address to which to bind all services",
+	Usage: "default IP address to which to bind shared services",
 }
 
 // TLS flags
--- a/cmd/tinkerbell/flag/smee.go
+++ b/cmd/tinkerbell/flag/smee.go
@@ -12,9 +12,11 @@ import (
 	"github.com/peterbourgon/ff/v4/ffval"
 	"github.com/tinkerbell/tinkerbell/pkg/backend/kube"
 	"github.com/tinkerbell/tinkerbell/pkg/constant"
+	"github.com/tinkerbell/tinkerbell/pkg/flag/delimitedlist"
 	ntip "github.com/tinkerbell/tinkerbell/pkg/flag/netip"
 	"github.com/tinkerbell/tinkerbell/pkg/flag/url"
 	"github.com/tinkerbell/tinkerbell/smee"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 type SmeeConfig struct {
@@ -25,7 +27,13 @@ type SmeeConfig struct {
 	// DHCPIPXEScript splits out some url.URL fields so they can be set individually.
 	// The cmd package is responsible for putting the fields back together into a url.URL for use in service package configs.
 	DHCPIPXEScript URLBuilder
-	LogLevel       int
+	// DHCPv6IPXEBinary splits out some url.URL fields so they can be set individually.
+	// The cmd package is responsible for putting the fields back together into a url.URL for use in service package configs.
+	DHCPv6IPXEBinary URLBuilder
+	// DHCPv6IPXEScript splits out some url.URL fields so they can be set individually.
+	// The cmd package is responsible for putting the fields back together into a url.URL for use in service package configs.
+	DHCPv6IPXEScript URLBuilder
+	LogLevel         int
 }
 
 var KubeIndexesSmee = map[kube.IndexType]kube.Index{
@@ -47,6 +55,29 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 	fs.Register(DHCPEnabled, ffval.NewValueDefault(&sc.Config.DHCP.Enabled, sc.Config.DHCP.Enabled))
 	fs.Register(DHCPEnableNetbootOptions, ffval.NewValueDefault(&sc.Config.DHCP.EnableNetbootOptions, sc.Config.DHCP.EnableNetbootOptions))
 	fs.Register(DHCPModeFlag, &sc.Config.DHCP.Mode)
+	fs.Register(DHCPv6Enabled, ffval.NewValueDefault(&sc.Config.DHCPv6.Enabled, sc.Config.DHCPv6.Enabled))
+	fs.Register(DHCPv6EnableNetbootOptions, ffval.NewValueDefault(&sc.Config.DHCPv6.EnableNetbootOptions, sc.Config.DHCPv6.EnableNetbootOptions))
+	fs.Register(DHCPv6ModeFlag, &sc.Config.DHCPv6.Mode)
+	fs.Register(DHCPv6DefaultNameServers, delimitedlist.NewParsed(&sc.Config.DHCPv6.DefaultNameServers, ',', parseDefaultIPv6NameServer))
+	fs.Register(DHCPv6DefaultDomainSearchList, delimitedlist.NewParsed(&sc.Config.DHCPv6.DefaultDomainSearchList, ',', parseDomainSearchSuffix))
+	fs.Register(DHCPv6ServerDUID, ffval.NewValueDefault(&sc.Config.DHCPv6.ServerDUID, sc.Config.DHCPv6.ServerDUID))
+	fs.Register(DHCPv6DerivedDirectAddressPool, &ntip.Prefix{Prefix: &sc.Config.DHCPv6.DerivedDirectAddressPool})
+	fs.Register(DHCPv6DerivedRelayAddressPrefix, ffval.NewValueDefault(&sc.Config.DHCPv6.DerivedRelayAddressPrefix, sc.Config.DHCPv6.DerivedRelayAddressPrefix))
+	fs.Register(DHCPv6BindAddr, &ntip.Addr{Addr: &sc.Config.DHCPv6.BindAddr})
+	fs.Register(DHCPv6BindPort, ffval.NewValueDefault(&sc.Config.DHCPv6.BindPort, sc.Config.DHCPv6.BindPort))
+	fs.Register(DHCPv6BindInterface, ffval.NewValueDefault(&sc.Config.DHCPv6.BindInterface, sc.Config.DHCPv6.BindInterface))
+	fs.Register(DHCPv6SyslogIP, &ntip.Addr{Addr: &sc.Config.DHCPv6.SyslogIP})
+	fs.Register(DHCPv6TftpIP, &ntip.Addr{Addr: &sc.Config.DHCPv6.TFTPIP})
+	fs.Register(DHCPv6TftpPort, ffval.NewValueDefault(&sc.Config.DHCPv6.TFTPPort, sc.Config.DHCPv6.TFTPPort))
+	fs.Register(DHCPv6IPXEHTTPScriptInjectMac, ffval.NewValueDefault(&sc.Config.DHCPv6.IPXEHTTPScript.InjectMacAddress, sc.Config.DHCPv6.IPXEHTTPScript.InjectMacAddress))
+	fs.Register(DHCPv6IPXEHTTPBinaryURLScheme, ffval.NewValueDefault(&sc.Config.DHCPv6.IPXEHTTPBinaryURL.Scheme, sc.Config.DHCPv6.IPXEHTTPBinaryURL.Scheme))
+	fs.Register(DHCPv6IPXEHTTPBinaryURLHost, ffval.NewValueDefault(&sc.DHCPv6IPXEBinary.Host, sc.DHCPv6IPXEBinary.Host))
+	fs.Register(DHCPv6IPXEHTTPBinaryURLPort, ffval.NewValueDefault(&sc.DHCPv6IPXEBinary.Port, sc.DHCPv6IPXEBinary.Port))
+	fs.Register(DHCPv6IPXEHTTPBinaryURLPath, ffval.NewValueDefault(&sc.Config.DHCPv6.IPXEHTTPBinaryURL.Path, sc.Config.DHCPv6.IPXEHTTPBinaryURL.Path))
+	fs.Register(DHCPv6IPXEHTTPScriptScheme, ffval.NewValueDefault(&sc.Config.DHCPv6.IPXEHTTPScript.URL.Scheme, sc.Config.DHCPv6.IPXEHTTPScript.URL.Scheme))
+	fs.Register(DHCPv6IPXEHTTPScriptHost, ffval.NewValueDefault(&sc.DHCPv6IPXEScript.Host, sc.DHCPv6IPXEScript.Host))
+	fs.Register(DHCPv6IPXEHTTPScriptPort, ffval.NewValueDefault(&sc.DHCPv6IPXEScript.Port, sc.DHCPv6IPXEScript.Port))
+	fs.Register(DHCPv6IPXEHTTPScriptPath, ffval.NewValueDefault(&sc.Config.DHCPv6.IPXEHTTPScript.URL.Path, sc.Config.DHCPv6.IPXEHTTPScript.URL.Path))
 	fs.Register(DHCPBindAddr, &ntip.Addr{Addr: &sc.Config.DHCP.BindAddr})
 	fs.Register(DHCPBindInterface, ffval.NewValueDefault(&sc.Config.DHCP.BindInterface, sc.Config.DHCP.BindInterface))
 	fs.Register(DHCPIPForPacket, &ntip.Addr{Addr: &sc.Config.DHCP.IPForPacket})
@@ -107,6 +138,7 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 	fs.Register(IPXEHTTPScriptRetries, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.Retries, sc.Config.IPXE.HTTPScriptServer.Retries))
 	fs.Register(IPXEHTTPScriptRetryDelay, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.RetryDelay, sc.Config.IPXE.HTTPScriptServer.RetryDelay))
 	fs.Register(IPXEHTTPScriptOSIEURL, &url.URL{URL: sc.Config.IPXE.HTTPScriptServer.OSIEURL})
+	fs.Register(IPXEHTTPScriptOSIEURLv6, &url.URL{URL: sc.Config.IPXE.HTTPScriptServer.OSIEURLv6})
 	fs.Register(IPXEScriptSyslogFQDN, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.SyslogFQDN, sc.Config.IPXE.HTTPScriptServer.SyslogFQDN))
 	fs.Register(IPXEBinaryInjectMacAddrFormat, &ffval.Enum[constant.MACFormat]{
 		ParseFunc: macAddrFormatParser,
@@ -117,6 +149,7 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 
 	// iPXE Tink Server Flags
 	fs.Register(TinkServerAddrPort, ffval.NewValueDefault(&sc.Config.TinkServer.AddrPort, sc.Config.TinkServer.AddrPort))
+	fs.Register(TinkServerAddrPortV6, ffval.NewValueDefault(&sc.Config.TinkServer.AddrPortV6, sc.Config.TinkServer.AddrPortV6))
 	fs.Register(TinkServerUseTLS, ffval.NewValueDefault(&sc.Config.TinkServer.UseTLS, sc.Config.TinkServer.UseTLS))
 	fs.Register(TinkServerInsecureTLS, ffval.NewValueDefault(&sc.Config.TinkServer.InsecureTLS, sc.Config.TinkServer.InsecureTLS))
 
@@ -149,47 +182,16 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 }
 
 // Convert CLI specific fields to smee.Config fields.
-func (s *SmeeConfig) Convert(trustedProxies *[]netip.Prefix, publicIP netip.Addr, bindAddr netip.Addr, defaultPort int) {
+func (s *SmeeConfig) Convert(trustedProxies *[]netip.Prefix, publicIP, publicIPv6 netip.Addr, bindAddr netip.Addr, defaultPort int) {
 	s.Config.IPXE.HTTPScriptServer.TrustedProxies = ntip.ToPrefixList(trustedProxies).Slice()
-	s.Config.DHCP.IPXEHTTPScript.URL.Host = func() string {
-		var addr string                        // Defaults
-		port := fmt.Sprintf("%d", defaultPort) // Defaults
-		if !publicIP.IsUnspecified() && publicIP.IsValid() {
-			addr = publicIP.String()
-		}
-		// CLI flag
-		if s.DHCPIPXEScript.Host != "" {
-			addr = s.DHCPIPXEScript.Host
-		}
-		if s.DHCPIPXEScript.Port != 0 {
-			port = fmt.Sprintf("%d", s.DHCPIPXEScript.Port)
-		}
-
-		if port != "" {
-			return fmt.Sprintf("%s:%s", addr, port)
-		}
-		return addr
-	}()
-
-	s.Config.DHCP.IPXEHTTPBinaryURL.Host = func() string {
-		var addr string                        // Defaults
-		port := fmt.Sprintf("%d", defaultPort) // Defaults
-		if !publicIP.IsUnspecified() && publicIP.IsValid() {
-			addr = publicIP.String()
-		}
-		// CLI flag
-		if s.DHCPIPXEBinary.Host != "" {
-			addr = s.DHCPIPXEBinary.Host
-		}
-		if s.DHCPIPXEBinary.Port != 0 {
-			port = fmt.Sprintf("%d", s.DHCPIPXEBinary.Port)
-		}
-
-		if port != "" {
-			return fmt.Sprintf("%s:%s", addr, port)
-		}
-		return addr
-	}()
+	s.Config.DHCP.IPXEHTTPScript.URL.Host = s.advertisedHost(s.DHCPIPXEScript, publicIP, defaultPort)
+	s.Config.DHCP.IPXEHTTPBinaryURL.Host = s.advertisedHost(s.DHCPIPXEBinary, publicIP, defaultPort)
+	if publicIPv6.IsValid() && !publicIPv6.IsUnspecified() || s.DHCPv6IPXEScript.Host != "" || s.DHCPv6IPXEScript.Port != 0 {
+		s.Config.DHCPv6.IPXEHTTPScript.URL.Host = s.advertisedHost(s.DHCPv6IPXEScript, publicIPv6, defaultPort)
+	}
+	if publicIPv6.IsValid() && !publicIPv6.IsUnspecified() || s.DHCPv6IPXEBinary.Host != "" || s.DHCPv6IPXEBinary.Port != 0 {
+		s.Config.DHCPv6.IPXEHTTPBinaryURL.Host = s.advertisedHost(s.DHCPv6IPXEBinary, publicIPv6, defaultPort)
+	}
 
 	// Service-specific bind addresses take precedence over the global bind address.
 	if bindAddr.IsValid() {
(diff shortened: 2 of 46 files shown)


brk0v:
Added a signature in the latest commit update.

cmd/tinkerbell/flag/smee.go
```diff
@@ -24,7 +25,13 @@ type SmeeConfig struct {
 	// DHCPIPXEScript splits out some url.URL fields so they can be set individually.
 	// The cmd package is responsible for putting the fields back together into a url.URL for use in service package configs.
 	DHCPIPXEScript URLBuilder
-	LogLevel       int
+	// DHCPv6IPXEBinary splits out some url.URL fields so they can be set individually.
+	// The cmd package is responsible for putting the fields back together into a url.URL for use in service package configs.
+	DHCPv6IPXEBinary URLBuilder
```
rahulbabu95, line 30: Since Ipxe doesn't have a differentiation between ipv4 & ipv6, is there something specific that necessitates a new field here for ipv6?

cmd/tinkerbell/flag/global.go
```diff
@@ -17,6 +17,8 @@ type GlobalConfig struct {
 	OTELInsecure         bool
 	TrustedProxies       []netip.Prefix
 	PublicIP             netip.Addr
+	PublicIPv6           netip.Addr
+	DualStack            bool
```
rahulbabu95: This was mentioned as an open question here: https://github.com/tinkerbell/roadmap/issues/44. Has this been tested thoroughly for smee to take the overhead/responsibility that comes in with this? Not against it, just trying to understand the confidence in additional scope.

docs/technical/DHCP_BOOT_MODES.md
```diff
@@ -44,3 +44,78 @@ To enable this mode set the CLI flag `--dhcp-enabled=false` or the environment v
...
+
+Derived mode has no lease database and does not track Duplicate Address Detection (DAD) failures. If a client receives a derived address, detects a duplicate address, and sends a DHCPv6 Decline, Tinkerbell replies according to the DHCPv6 Decline flow but does not mark the address as unusable or choose a replacement. Because derived addresses are deterministic from the client MAC address and selected prefix, the same client will receive the same derived address again until the duplicate is removed or the derived prefix configuration changes.
+
+Direct client requests use `--dhcpv6-derived-direct-address-pool` / `TINKERBELL_DHCPV6_DERIVED_DIRECT_ADDRESS_POOL`. The prefix must be a usable IPv6 unicast prefix between `/1` and `/64`; `/65` through `/128` are rejected because they do not leave enough host bits for stateless MAC-based derivation, and unusable ranges such as unspecified, IPv4-mapped, link-local, or multicast prefixes are rejected. If no direct pool is configured, derived mode ignores direct requests that do not already have a Hardware IPv6 reservation.
```
rahulbabu95: nit: But when i tried testing with this flag, i set this to a different pool than RA prefix and the boot-time address was unroutable which made provisioning fail at netboot during TFTP. Do you think it's worth calling out here that the derived pool must be on-link for the provisioning segment (since we don't manage RA)?

cmd/tinkerbell/flag/smee.go
brk0v: Thank you for your time and the review. 

My intention was to keep the IPv4 and IPv6 paths independent. 
I think about a use case with an environment that requires both IPv4 and IPv6 DHCP servers enabled: some hosts support IPv6 while others do not. 
In that situation, I would prefer to configure all URLs using IP addresses rather than domain names, since DNS and resolver configuration can introduce additional complexity in mixed-stack setups.

docs/technical/DHCP_BOOT_MODES.md
brk0v: Thank you for your time and the review. 
Agreed, updated the documetation.

cmd/tinkerbell/flag/global.go
brk0v: Thank you for your time and the review.

After some additional thought and iteration, I removed the DualStack field. The flag only controlled the default bind address for shared services, and it did not provide guaranteed end-to-end dual-stack behavior. R

The default bind-address selection remains IPv4-first and backward compatible when an IPv4 address is available. Admins can explicitly set `--bind-address=::` to bind shared services to the IPv6 wildcard.

Within Smee, DHCPv4 and DHCPv6 remain separate listeners with independent configuration and enablement, and DHCPv6 is disabled by default.

cmd/tinkerbell/flag/tink_server.go
```diff
@@ -33,9 +33,9 @@ func RegisterTinkServerFlags(fs *Set, t *TinkServerConfig) {
 
 // Convert TinkServerConfig data types to tink server server.Config data types.
 func (t *TinkServerConfig) Convert(bindAddr netip.Addr) {
-	addr := t.BindAddr
-	if bindAddr.IsValid() {
```
brk0v: I think here was a bug. The incoming `bindAddr` comes from global and always valid. The logic should be inverted.

brk0v: Fixing it in dedicated PR https://github.com/tinkerbell/tinkerbell/pull/898

brk0v: fixed

jacobweinstock:
Hey @brk0v. This is an AI review to get started. I have a human one in coming too. Thanks!

## Review: IPv6/DHCPv6 support — correctness, RFC conformance, compat, security, design

**TL;DR:** High-quality protocol work, strong tests, unusually honest docs. I verified the claims below rather than assuming them. There is **1 backward-compat regression**, one RFC MUST violation, and **one structural design defect** (no construction-time invariants) that is both the root cause of a real bug in this PR (D1/O2 below) and the thing that will make this code expensive to change. Everything else can be follow-up issues.

Items are numbered (**C**orrectness, **R**FC, **D**esign, **O**perational, **B**ackward-compat) for easy reference in replies.

---

### ✅ Requested before merge

- [ ] **C1** 🔴 Fix `Is4In6` regression in `hasIPv4Reservation` — IPv4-mapped reservations (`::ffff:a.b.c.d`) silently stop working on DHCPv4 ([handler.go#L252](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/handler/reservation/handler.go#L252-L254))
- [ ] **R1** 🔴 Clamp Information Refresh Time to IRT_MINIMUM (600s) — RFC 9915 §21.23 MUST ([stateless.go#L260](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/stateless/stateless.go#L260-L268))
- [ ] **D1** 🔴 Fail at startup, not per packet, when DHCPv6 netboot is enabled but unconfigurable (today clients silently receive `http:///…` boot URLs). Root cause and recommended shape in **Design** below — a validating constructor makes this bug unrepresentable rather than re-checked.
- [ ] **O1** 🟡 Shared hardcoded fallback server DUID — at minimum warn at startup ([smee.go#L89](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/smee.go#L89-L92))
- [ ] **O2** 🟡 iPXE syslog family: key on script family (`auto.ipxe` vs `auto6.ipxe`), not DNS AAAA presence — dual-stack `SyslogFQDN` flips existing IPv4 clients to `set syslog6` ([hook.go#L98](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/ipxe/script/hook.go#L98-L136))

Suggested follow-up issues (not blocking): R2–R4, D2–D4, O3–O4, B1–B4.

---

<details>
<summary><b>What I validated (proven, not assumed)</b></summary>

| Check | Result |
|---|---|
| `go build ./...`, `go vet ./...` | ✅ clean |
| `go test` across smee, cmd, pkg, secondstar | ✅ all pass |
| `go test -race -count=2` on all 4 new dhcpv6 packages | ✅ pass |
| `make helm-lint` and `make helm-template` (incl. negative assertions) | ✅ exit 0 |
| RFC references (`rfc9915#section-21.16`, etc.) | ✅ correct — [RFC 9915](https://www.rfc-editor.org/rfc/rfc9915.html) (Jan 2026) obsoletes RFC 8415 |
| `ClientEnterpriseNumber = 343` "works with EDK II" | ✅ verified in [tianocore/edk2 `PxeBcDhcp6.h`](https://github.com/tianocore/edk2/blob/master/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.h): `#define PXEBC_DHCP6_ENTERPRISE_NUM 343` |
| Rapid Commit (§21.14 MUST include in Reply) | ✅ library's `NewReplyFromMessage` auto-prepends `WithRapidCommit` |
| Multicast `ff02::1:2` / `ff05::1:3` joined | ✅ library joins both when bound to wildcard `[::]:547` (the default). Caveat: a custom **unicast** `--dhcpv6-bind-addr` silently disables multicast reception — worth a doc note |
| `--dhcpv6-bind-interface` comma-list | ✅ feasible: lib sets `SO_REUSEADDR` + `SO_BINDTODEVICE` (needs CAP_NET_RAW; lib errors helpfully without it) |
| Docs accuracy (DHCP_BOOT_MODES.md, PORTS_AND_ENDPOINTS.md) | ✅ match the code I traced, incl. DAD/Decline limitation, derived-pool on-link requirement, T1/T2 policy, bind-address table |
| Defaults / backward compat | ✅ DHCPv6 off by default everywhere; bind-address default remains IPv4-first ([defaults.go#L225](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/cmd/tinkerbell/defaults.go#L225-L233)) |

</details>

<details>
<summary><b>C1 — Backward-compat regression: IPv4-mapped reservations (must fix)</b></summary>

[`hasIPv4Reservation`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/handler/reservation/handler.go#L252-L254) checks only `Is4()`, but the unchanged [conv.go](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/conv.go#L118-L135) explicitly treats `Is4In6()` (`::ffff:a.b.c.d`) as an IPv4 reservation (netmask required, broadcast computed), and `dhcpv4.WithYourIP(d.IPAddress.AsSlice())` handled the mapped form. Such Hardware worked before this PR; now DISCOVER/REQUEST are silently ignored.

Fix: `d.IPAddress.Is4() || d.IPAddress.Is4In6()`. (The gate itself is otherwise a good fix — previously an IPv6 reservation produced a garbage `yiaddr` offer.)

</details>

<details>
<summary><b>D1–D4 — Design: invariants, coupling, and why bug D1 exists</b></summary>

**D1 — Root cause: no handler establishes invariants at construction.** Handlers are assembled as bare struct literals with exported fields, so validity is never guaranteed and the cost is paid in three places:

1. *Upward coupling.* [`dhcpv6Handler`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/smee.go#L779-L883) (105 lines) owns the handler packages' validity rules — TFTP addr/port legality, derived-pool bounds, and the convention that "netboot disabled" means *leave three dependencies nil/zero but set `Enabled: false` so the handler knows to skip them*. That knowledge belongs to the handler packages. Today, adding one netboot field is a 5-site edit (4 switch arms + the per-packet copy in D2), and getting the nil-choreography wrong compiles cleanly.
2. *Per-packet defensive re-validation.* Because nothing guarantees the receiver is valid, [`Handle` re-checks `h.Backend == nil || h.ServerID == nil` on every packet](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/reservation/reservation.go#L106-L115) (stateless too), and [`Info.BootURL`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/dhcpv6.go#L246-L305) re-discovers *configuration* problems ("missing TFTP iPXE binary server", "missing iPXE script URL builder") in the packet hot path.
3. *Config errors demoted to runtime errors.* Enabling DHCPv6 netboot without an advertised IPv6 host passes startup ([`ipxeScriptURLBuilder`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/smee.go#L957-L982) accepts `http:///ipxe/script/auto6.ipxe`) and surfaces as per-client boot failures. This is a shipping bug in the PR, and it is a direct consequence of 1+2.

*Recommended shape:* each handler package exports `New(Config) (*Handler, error)` that validates once; replace the `Netboot.Enabled bool` + nil-fields pattern with a small consumer-defined interface (e.g. `bootURLSource`) injected by `smee.go` — a validated real implementation when netboot is on, a no-op when off. `Handle` then contains zero self-validation, `dhcpv6Handler` collapses to a small mode→constructor table, and D1's bug class becomes unrepresentable. (This is Cheney's SOLID-for-Go directly: validity knowledge lives with the type — SRP; behavior varies by substituting an implementation, not by flag-guarded branches — OCP/LSP; the handler depends on the smallest interface it defines — ISP/DIP.) Note this pattern is inherited from the DHCPv4 `dhcpHandler` — the PR copies the house style — but this PR doubles its surface area, which is the right moment to stop.

**D2 —** [`reservation.Handler.statelessHandler()`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/reservation/reservation.go#L280-L298) hand-builds a fifth copy of the shared config **per Information-Request packet**, and couples the reservation package to the concrete stateless package. Compose the delegation once in `smee.go` or inject an interface. Relatedly, `reservation.Netboot`/`stateless.Netboot` are identical copied structs converted field-by-field — the "little copying" proverb doesn't apply here because the copies don't evolve independently; they're manually synced mirrors.

**D3 —** In [ipxe.go](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/ipxe/script/ipxe.go#L117-L127), the same `if isIPv6 {…}` three-field swap is duplicated in `serveStaticIPXEScript` and `defaultScript`, with an `isIPv6 bool` threaded through three call layers — a boolean parameter standing in for a missing per-family view type (`settingsFor(family)`).

**D4 —** Smaller readability items: [Convert](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/cmd/tinkerbell/flag/smee.go#L186-L194) gates twice on unparenthesized `a && b || c || d`; `advertisedHost` keeps a leftover `return func() string {…}()` IIFE; the Makefile's 22 `helm template … | grep -q` one-liners are an unnamed, undiagnosable test suite in the wrong medium (negative `!` cases pass when templating fails for the *wrong* reason) — [helm-unittest](https://github.com/helm-unittest/helm-unittest), a case-table script, or a Go test over parsed YAML would give named, reportable cases.

*Credit where due:* consumer-defined `BackendReader` interfaces per package, the one-method `serverv6.Handler` interface, and `func(net.HardwareAddr) *url.URL` instead of a one-method interface are exactly right.

</details>

<details>
<summary><b>R1–R4 — RFC 9915 conformance (low impact, cheap to fix)</b></summary>

- **R1** (🔴, in checklist): [`informationRefreshTime`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/stateless/stateless.go#L260-L268) only guards `<= 0` and `> max`; [§21.23](https://www.rfc-editor.org/rfc/rfc9915.html#section-21.23) says the value **MUST NOT** be below IRT_MINIMUM (600s).
- **R2**: §16 validation gaps — Solicit containing our own ServerID accepted (§16.2: MUST discard *any* Solicit with a ServerID); same for Rebind (§16.7); Info-Request containing IA options not rejected (§16.12).
- **R3**: T1/T2 policy in [`statefulLifetimes`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/reservation/reservation.go#L422-L436): preferred = valid/2, T1 = preferred, T2 = 1.6·preferred vs recommended 0.5/0.8 × preferred — renew starts exactly at deprecation. Legal (SHOULD), documented, but unusual.
- **R4**: SHA-derived and fallback DUID-UUIDs don't set RFC 4122 version/variant bits (RFC 6355). Harmless — DUIDs are opaque per §11 — just untidy.

</details>

<details>
<summary><b>O1–O4 — Operational risks</b></summary>

- **O1** (🟡, in checklist): [fallback DUID](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/smee.go#L89-L92) is identical across all deployments without config — two Smee instances sharing an L2/relay scope claim the same server identity. The [auto-derived DUID](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/smee.go#L903-L925) also changes when `TinkServer.AddrPortV6` changes, invalidating bindings mid-lease.
- **O2** (🟡, in checklist): [`SyslogIPXEConfigName`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/ipxe/script/hook.go#L98-L136) prefers `syslog6` when the shared `SyslogFQDN` has *any* AAAA record — IPv4-booted clients on dual-stack DNS lose iPXE console logging. Also a blocking, uncached `net.LookupHost` in the script-render path.
- **O3**: Reservation-mode Info-Request parity gap — the address gate runs *before* delegation to the stateless handler ([reservation.go#L190](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/handler/reservation/reservation.go#L190-L212)), so hardware without an IPv6 reservation gets Info-Requests dropped. Documented policy; flagging for a conscious decision.
- **O4** (minor): relay-path `extractMAC` logs a stale nil `err` ([dhcpv6.go#L595](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/dhcpv6.go#L595-L601)); `DHCPv6.server` written in `Serve`/read in [`Close`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcpv6/server/dhcpv6.go#L193-L199) without sync (safe today, latent); [`parseDomainSearchSuffix`](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/cmd/tinkerbell/flag/smee.go#L307-L318) rejects trailing-dot FQDNs.

</details>

<details>
<summary><b>B1–B4 — Other backward-compat deltas (low-risk, but real)</b></summary>

- **B1**: DHCPv4 `Bootfile` refactor ([dhcp.go#L348](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/dhcp/dhcp.go#L348-L385)): HTTPClient case with nil HTTP bin server now yields `""` (was `/no-ipxe-script-defined`); iPXE-userclass with invalid TFTP AddrPort/empty binary now yields `""` (was a garbage `tftp://` URL). Misconfig-only, arguably better — but a v4 wire-behavior change riding along in an IPv6 PR.
- **B2**: Syslog receiver `udp4` → `udp` ([receiver.go#L34](https://github.com/tinkerbell/tinkerbell/blob/9167b07cb5c6037be9d8cce70f6c27a5d92a7242/smee/internal/syslog/receiver.go#L34-L40)) — a `:514`-style bind is now dual-stack.
- **B3**: `--public-ipv4` now hard-errors on non-IPv4 values (previously tolerated). Good, but a new startup failure mode.
- **B4**: Helm — the main Service now auto-emits `kube-vip.io/loadbalancerIPs` for existing IPv4 installs (benign for kube-vip, ignored by other LBs). The osie Service gained dual-IP annotation logic but not the `ipFamilyPolicy`/`ipFamilies` knobs the main Service got.

</details>

<details>
<summary><b>Security review</b></summary>

No injection or memory-safety issues found. Input parsing (`archFromDHCPv6VendorClass`, MAC extraction, DUID parsing) is bounded and error-checked; `notAllowedBootURL` strips query/fragment; message-type whitelisting and server-ID matching are solid. DoS posture: bounded worker pool with drop-on-full-queue and 10s per-packet handler timeout is reasonable; per-client rate limiting is out of scope per §14.1. Derived addresses are deterministic (deviation from §13.1 SHOULD NOT / RFC 7721) — deliberate, boot-time-only, documented; the subnet-router-anycast dodge in `derivedAddressFromHash` is a nice touch. Security-adjacent items: O1 (shared DUID) and the disclosed derived-address determinism.

</details>

---

**Bottom line:** merge-worthy after C1, R1, and D1 (which also fixes the silent-broken-boot-URL bug at its root rather than patching it). D2/D3 are the highest-leverage follow-ups — the config copies are already at 5× and every future field pays the tax. Happy to file the follow-up issues for R/D/O/B items.

jacobweinstock:
Hey @brk0v. Update for you. I've setup the tinkerbell playground for ipv6 only and have been testing this PR. Looking good so far. I will have a final review completed shorting. Thanks for all the work on this!

image

jacobweinstock:
Hey @brk0v. Here's the next round. I found these issues while testing with the Playground.

## IPv6 gaps: Smee ISO patching and tink-agent container networking

**TL;DR:** Four IPv6-only defects, two are **hard failures** (isoboot never gets a
Tink address; action containers get no usable network), two are **silent
degradation** (a malformed `syslog_host` token, and DNS that eats two resolver
timeouts per lookup). They compound: C2 and C3 each mask the other's symptoms,
so fixing one in isolation still looks broken.

Items are numbered (**C**orrectness, **O**perational) for easy reference in
replies.

---

### ✅ Requested before merge

- [ ] **C2** 🔴 ISO patching has no IPv6 path — `grpc_authority=` is emitted
      empty on IPv6-only, agent restart-loops forever
      ([iso.go#L353](smee/internal/iso/iso.go#L353))
- [ ] **C3** 🔴 Fallback CNI bridge is hardcoded IPv4 — action containers get no
      network on an IPv6-only host
      ([containerd.go#L68](tink/agent/internal/runtime/containerd/containerd.go#L68))
- [ ] **C1** 🟡 `syslogHost()` renders `netip.Addr`'s `"invalid IP"` sentinel
      into the cmdline and into an iPXE `set` line
      ([smee.go#L519](smee/smee.go#L519))
- [ ] **O1** 🟡 Fallback nameservers exceed glibc `MAXNS` — both usable entries
      wasted on IPv6-only
      ([dns.go#L48](tink/agent/internal/runtime/containerd/dns.go#L48))

C1 and C2 are independent but appear together in the same broken cmdline. O1
depends on whatever host family detection C3 settles on.

---

<details>
<summary><b>C1 — <code>syslogHost()</code> renders the literal string <code>invalid IP</code></b></summary>

### Problem

`Config.syslogHost()` ends in an unguarded `c.DHCP.SyslogIP.String()`. Go's
`netip.Addr.String()` returns the literal `"invalid IP"` for the zero value, so
with no IPv4 syslog address configured we emit `syslog_host=invalid IP`.

`SmeeConfig.Convert()` only populates `DHCP.SyslogIP` when `publicIP` is valid
and non-unspecified, so this hits every IPv6-only deployment.

Two consumers are affected:

- **ISO cmdline** (`iso.KernelParams.Syslog`) — lands verbatim on the kernel
  command line.
- **iPXE script** (`Handler.PublicSyslogFQDN`) — worse, because it goes into
  `set syslog-address:ipv4 {{ .SyslogHost }}`. The embedded space makes iPXE
  parse `IP` as a stray argument, so the script line itself is malformed.

`syslogHostV6()` already guards against invalid/unspecified/wrong-family
addresses. `syslogHost()` never got the same treatment.

### Fix options

1. Guard in `syslogHost()`, mirroring `syslogHostV6()` exactly: return `""` for
   an invalid, unspecified, or non-IPv4 address. `DHCP.SyslogIP` is IPv4-only by
   design — IPv6 has its own `DHCPv6.SyslogIP` — so rejecting a v6 value there
   is correct rather than a compatibility risk. Downstream already handles
   empty: the iPXE template is wrapped in `{{- if .SelectedSyslogHost }}`, and
   an empty `syslog_host=` is inert.
2. Guard at each call site — the same bug returns with the next caller.
3. Validate at config load and fail startup. Arguably most correct, but it
   breaks IPv6-only deployments that are otherwise working today.

</details>

<details>
<summary><b>C2 — ISO patching is IPv4-only, so isoboot on IPv6 is silently broken</b></summary>

### Problem

`iso.KernelParams` has no IPv6 members, and `Config.ISOHandler()` wires only the
v4 fields:

```go
KernelParams: iso.KernelParams{
    Syslog:             c.syslogHost(),        // v4 only
    TinkServerGRPCAddr: c.TinkServer.AddrPort, // v4 only
}
```

`ScriptHandler()`, immediately above it, does wire both families via
`PublicSyslogFQDNV6` / `TinkServerGRPCAddrV6`. `constructPatch()` has no
address-family branch at all.

On an IPv6-only deployment `TinkServer.AddrPort` is never populated (`Convert()`
gates it on a valid `publicIP`), so the patched ISO boots with:

```
syslog_host=invalid IP grpc_authority= tinkerbell_tls=false worker_id=...
```

The agent then fails every start with `unable to create gRPC client: the
Tinkerbell server address is required, none provided` and systemd
restart-loops. The correctly-derived `TinkServer.AddrPortV6` and
`DHCPv6.SyslogIP` are in the config, just never read by this path.

Net effect: netboot works on v4 and v6, isoboot works on v4, isoboot on v6 has
no code path — and nothing reports a misconfiguration, so it presents as an
agent bug on the machine rather than a Smee one.

### Fix options

Adding `SyslogV6` / `TinkServerGRPCAddrV6` to `KernelParams` is
uncontroversial. The design question is how `constructPatch` picks a family,
since there is no `auto.ipxe` vs `auto6.ipxe` signal here.

1. **Hardware record's assigned address** (`d.IPAddress`). `constructPatch`
   already receives the `*dhcp.DHCP`, and it is the same source that drives the
   `ipam=` parameter, so the cmdline stays internally consistent. Describes what
   the machine can reach once booted.
2. **Inbound request's address family** (`req.RemoteAddr`). Describes how the
   ISO was *fetched*, not how the booted OS reaches Tink, and is unreliable
   behind the LB/trusted-proxy setup.
3. **Separate URL path** (`/iso6/<mac>/...`), mirroring `auto6.ipxe`. Most
   explicit, but a user-visible change needing the Helm chart, `isoURL` in
   Hardware CRs, and external tooling updated in lockstep.
4. **Fall back to the v6 config when v4 is absent.** Least code, but papers over
   the mismatch and produces a wrong cmdline on genuine dual-stack.

The `invalid IP` half of that cmdline is C1.

</details>

<details>
<summary><b>C3 — Fallback CNI bridge is hardcoded IPv4, leaving action containers with no network</b></summary>

### Problem

`fallbackBridgeConflist` is a `const` with one hardcoded IPAM range
(`172.17.0.0/23`) and one route (`0.0.0.0/0`). This is what the agent
synthesises when `/etc/cni/net.d/` has no usable conflist — the default path on
a stock CaptainOS/HookOS boot.

On an IPv6-only host every action container gets an RFC1918 address, a default
route to a gateway with no uplink, and no IPv6 address. Image pulls, artifact
fetches and Tink itself have no path out. It surfaces inside the action as a
generic dial/timeout error, far from the CNI config that caused it.

Always emitting dual-stack is not the fix: a default route for a family with no
uplink means connections on that family stall until timeout rather than failing
fast. The conflist has to match the host.

Related, same area: `isIPForwardingEnabled` only reads
`/proc/sys/net/ipv4/ip_forward`. On an IPv6-only host that logs "IP forwarding
is disabled, CNI bridge NAT will not work" — a red herring — while never
checking `/proc/sys/net/ipv6/conf/all/forwarding`, which is the one that
matters.

### Fix options

How should the agent decide which families to configure?

1. **Default-route presence**, from `/proc/net/route` and
   `/proc/net/ipv6_route`. Strongest available proxy for "this family has an
   uplink", no network I/O, cheap at startup. Would need to skip `RTF_REJECT`
   entries, since the kernel installs an unreachable default in some configs.
2. **Address presence** via `net.Interfaces()`. Obvious first instinct and
   actively wrong: the `cni0` bridge the agent creates carries a global address
   in whichever families the conflist declared, so after one run a family looks
   reachable when it goes nowhere. Only misbehaves on the second and subsequent
   boots.
3. **Dual-stack unconditionally.** Rejected for the stall behaviour above.
4. **Agent flag / kernel cmdline parameter.** Pushes the burden onto the
   operator for something the agent can observe directly. Possibly worth having
   as an override regardless.
5. **Active probe** per family. Most accurate, but adds startup latency and a
   network dependency, and fails closed in air-gapped deployments.

Whatever is chosen, two details need settling: what to do when neither family is
detected (keeping IPv4 preserves container-to-container networking rather than
emitting an empty `ranges`), and which ULA prefix to use for the v6 range.

Scope note: this is the **fallback** path only. Configs in `/etc/cni/net.d/`,
HookOS- or user-supplied, still win untouched.

</details>

<details>
<summary><b>O1 — Fallback nameservers exceed glibc's 3-resolver limit</b></summary>

### Problem

```go
var fallbackNameservers = []string{"8.8.8.8", "8.8.4.4", "2001:4860:4860::8888", "2001:4860:4860::8844"}
```

This is written verbatim into the container's `/etc/resolv.conf` when the host
has no usable nameservers — `/etc/resolv.conf` unreadable, or every entry a
localhost stub that means nothing inside the container's netns.

**glibc only honours the first three `nameserver` lines** (`MAXNS == 3`);
anything past that is silently discarded. On an IPv6-only host that gives two
unreachable IPv4 resolvers, one usable IPv6 resolver, and one usable IPv6
resolver that is never read.

It doesn't fail outright, which is what makes it hard to spot: every lookup
burns two resolver timeouts before reaching the entry that works. Image pulls
and network-facing actions appear to hang, then eventually succeed, with
nothing in the logs pointing at DNS.

Distinct from C3, but the same root shape, and they mask each other — fixing the
bridge alone leaves DNS crawling.

### Fix options

1. **Select the list by address family**, sharing whatever host family
   detection C3 lands on. Dual-stack gets three entries that fit inside MAXNS
   and still cover both families; single-family hosts get two working resolvers
   and nothing wasted.
2. **Reorder so IPv6 comes first.** Cheaper, but only moves the problem to
   IPv4-only hosts, and one entry is still silently dropped.
3. **Truncate to three** without family awareness. Fixes the silent drop but not
   the unreachable-first ordering, which is where the latency comes from.
4. **Probe each resolver at startup.** Same objections as in C3.
5. **Make the fallback resolvers configurable.** Worth doing independently —
   hardcoding Google Public DNS is its own discussion for air-gapped and
   privacy-sensitive deployments — but a configured list has the same MAXNS
   problem, so it isn't a substitute for (1).

If (1) is chosen, the dual-stack ordering is a bikeshed worth settling:
`[v4, v6, v4]` favours a healthy v4 path, `[v4, v6, v6]` is equally defensible.

</details>

jacobweinstock (approved):
Thanks, @brk0v !

brk0v:
@jacobweinstock thank you!

PR #847, opened 2026-07-10
Support interleaving webhook calls with BMC actions in customboot

mcanevet:
## Summary
- `preparingActions`/`postActions` in `customboot` now accept a heterogeneous list of entries: a BMC action (unchanged shape, e.g. `- powerAction: "off"`) or a webhook call (`- webhook: {url: ...}`), in any order.
- Webhook entries are called directly by the Tink Controller; contiguous runs of BMC-action entries are still batched into a single `bmc.Job`, so an existing pure-BMC-actions list behaves exactly as it always has.
- Progress through each list is tracked in the new `status.bootOptions.actions` field.
- `webhook` supports templated `url`/`body`/static header values, Secret-backed headers (`valueFrom`), and HTTP Basic auth via a Secret reference (`basicAuth`), matching the existing BMC-credential convention.
- Failures in hardware lookup / templating / webhook calls during a customboot phase now set a `BootJobSetupFailed` condition, matching the existing netboot/isoboot diagnostics.

See `docs/technical/BOOT_MODES.md` (Webhooks in customboot) for usage and examples.

## Compatibility
- Additive/in-place retyping of `preparingActions`/`postActions`; a Workflow with no `webhook` entries is unaffected.
- Batch `bmc.Job` naming changes slightly (now suffixed by batch start index) for `customboot` phases.
- A Workflow that fully completed its `postActions` before upgrading will re-run `postActions` once on the first reconcile after upgrading, since the new progress-tracking status field has no record of prior completion.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean (root module and `api` submodule)
- [x] `go test ./...` passing (root module and `api` submodule)
- [x] `golangci-lint run` clean (only pre-existing, unrelated `goconst` findings)
- [x] Verified end-to-end against a real cluster and real hardware: webhook call (network-unreachable failure, non-2xx failure, and 2xx success), interleaved with real `powerAction`/`bootDevice` BMC actions executed against a real BMC via the `gofish` provider


jacobweinstock:
Hey @mcanevet. Thanks for this.

I thinking about how this fits into the v1alpha2 spec. I think it is unneeded in the v1alpha2 spec because the Workflow object allows for multiple tasks that each have their own bootOptions. So instead of having a webhook built into Tinkerbell, and having to deal with the code, the auth, etc, the first Task in the Workflow would be the "webhook". Then the another Task would handle booting the machine and provisioning it. This is quite a bit more flexible too, then just a single webhook. And as the example shows it can handle things like waiting for certain operations to complete, a vlan change on the switch for example. This isn't possible with just a webhook that then immediately moves on to networking booting a machine. There's a potential for race conditions.

The challenge is that I don't see a way to get this into the v1alpha1 spec. This is just brainstorming and I'd love to have your thoughts here.

Here's an example with the v1alpha2 specs:
```yaml
# Task 1: the "webhook".
# Instead of a webhook being built into Tinkerbell, the first Task in the
# Workflow acts as the webhook. It runs a single Action that calls out to an
# external system (IPAM, CMDB, notification endpoint, approval gate, etc.).
# This Task does not boot the machine, it just needs an Agent to execute the
# HTTP call. Because it runs before the provisioning Task, it can gate or
# enrich provisioning based on the external system's response.
apiVersion: tinkerbell.org/v1alpha2
kind: Task
metadata:
  name: webhook
  namespace: tinkerbell
spec:
  actions:
    - name: "notify-external-system"
      image: "curlimages/curl:8.11.0"
      command: "/bin/sh"
      args:
        - "-c"
        # Hardware templating lets the webhook payload carry machine details.
        - |
          curl --fail --silent --show-error \
            --request POST \
            --header "Content-Type: application/json" \
            --data '{"event":"provisioning-started","mac":"{{ .hardware.spec.network.interfaces.0.dhcp.mac }}","hostname":"{{ .hardware.spec.metadata.instance.hostname }}","vlan":"provisioning"}' \
            "{{ .references.webhookConfig.data.url }}"
      timeoutSeconds: 60
      retries: 3
  references:
    "webhookConfig":
      name: "webhook-settings"
      namespace: "tinkerbell"
      resource: "configmaps"
      version: "v1"
```

```yaml
# Task 2: wait for the external system to finish.
# After the webhook has kicked off an asynchronous change in an external system
# (for example, a VLAN change on the switch fabric), this Task polls that system
# until it reports the expected value. This gates the provisioning Task from
# starting until the network change has completed.
apiVersion: tinkerbell.org/v1alpha2
kind: Task
metadata:
  name: wait-for-vlan
  namespace: tinkerbell
spec:
  actions:
    - name: "poll-until-ready"
      image: "curlimages/curl:8.11.0"
      command: "/bin/sh"
      args:
        - "-c"
        # Poll the external system until it reports the VLAN change is complete.
        # The endpoint and the value we are waiting for come from a ConfigMap.
        - |
          set -eu
          url="{{ .references.vlanConfig.data.statusURL }}"
          want="{{ .references.vlanConfig.data.expectedState }}"
          echo "Waiting for '${want}' from ${url}"
          while true; do
            got="$(curl --fail --silent --show-error "${url}" || true)"
            echo "current state: '${got}'"
            if [ "${got}" = "${want}" ]; then
              echo "VLAN change complete"
              break
            fi
            sleep 10
          done
      # Bound the total polling time; the Action fails if the value never appears.
      timeoutSeconds: 900
  references:
    "vlanConfig":
      name: "vlan-settings"
      namespace: "tinkerbell"
      resource: "configmaps"
      version: "v1"
```

```yaml
# Task 3: boot and provision the machine.
# This is the "real" provisioning Task. It has its own bootOptions on the
# Workflow (toggleNetboot) so the machine is netbooted only for this Task, and
# it runs the Actions that stream an OS image to disk and reboot into it.
apiVersion: tinkerbell.org/v1alpha2
kind: Task
metadata:
  name: provision
  namespace: tinkerbell
spec:
  actions:
    - name: "stream-image"
      image: "quay.io/tinkerbell/actions/image2disk:latest"
      timeoutSeconds: 600
      envVars:
        - key: "DEST_DISK"
          value: "/dev/sda"
        - key: "IMG_URL"
          value: "{{ .references.osConfig.data.imageURL }}"
        - key: "COMPRESSED"
          value: "true"
    - name: "reboot"
      image: "quay.io/tinkerbell/actions/reboot:latest"
      timeoutSeconds: 90
      # Run in the background so the reboot does not block the Action reporting
      # success before the machine goes down.
      background: true
      namespaces:
        pid: "host"
  references:
    "osConfig":
      name: "os-settings"
      namespace: "tinkerbell"
      resource: "configmaps"
      version: "v1"
```

```yaml
# Task 4: the closing "webhook".
# After the machine has been provisioned on the provisioning VLAN, this Task
# signals the external system to move the machine back into the production VLAN.
# Like the first webhook, it only needs an Agent to make the HTTP call and does
# not boot the machine.
apiVersion: tinkerbell.org/v1alpha2
kind: Task
metadata:
  name: webhook-production-vlan
  namespace: tinkerbell
spec:
  actions:
    - name: "move-to-production-vlan"
      image: "curlimages/curl:8.11.0"
      command: "/bin/sh"
      args:
        - "-c"
        # Signal the external system to move the machine into the production VLAN.
        - |
          curl --fail --silent --show-error \
            --request POST \
            --header "Content-Type: application/json" \
            --data '{"event":"provisioning-complete","mac":"{{ .hardware.spec.network.interfaces.0.dhcp.mac }}","vlan":"production"}' \
            "{{ .references.webhookConfig.data.url }}"
      timeoutSeconds: 60
      retries: 3
  references:
    "webhookConfig":
      name: "webhook-settings"
      namespace: "tinkerbell"
      resource: "configmaps"
      version: "v1"
```

```yaml
# The Workflow ties the Tasks together in order.
# Task 1 (webhook) runs first, then Task 2 (wait-for-vlan) polls the external
# system until the change completes, then Task 3 (provision) boots and
# provisions the machine, then Task 4 (webhook-production-vlan) signals the
# external system to move the machine back into the production VLAN. Each Task
# has its own bootOptions, so only the provisioning Task netboots the machine.
apiVersion: tinkerbell.org/v1alpha2
kind: Workflow
metadata:
  name: webhook-then-provision
  namespace: tinkerbell
spec:
  timeoutSeconds: 1800
  tasks:
    # First Task: the webhook. No bootOptions are needed, it only calls the
    # external system. It still needs an Agent to execute the Action.
    - agentID: "abcd"
      taskRef:
        name: "webhook"
        namespace: "tinkerbell"
      timeoutSeconds: 120
    # Second Task: poll the external system until the VLAN change completes.
    # No bootOptions, it only waits on the external system.
    - agentID: "abcd"
      taskRef:
        name: "wait-for-vlan"
        namespace: "tinkerbell"
      timeoutSeconds: 900
    # Third Task: boot the machine and provision it. This Task carries the
    # bootOptions so only this Task netboots the Hardware.
    - agentID: "1234"
      taskRef:
        name: "provision"
        namespace: "tinkerbell"
      hardware:
        bootOptions:
          toggleNetboot: true
          bootMode: netboot
        hardwareRef:
          name: "example"
          namespace: "tinkerbell"
      timeoutSeconds: 1200
    # Fourth Task: the closing webhook. No bootOptions, it only signals the
    # external system to move the machine into the production VLAN.
    - agentID: "abcd"
      taskRef:
        name: "webhook-production-vlan"
        namespace: "tinkerbell"
      timeoutSeconds: 120
```

mcanevet:
Thanks for the detailed writeup, @jacobweinstock — this is a great thought experiment. A few reactions after digging into it:

**What is `abcd`?**

Could you clarify what you have in mind for the Agent behind `agentID: "abcd"` in Task 1? Looking at how Agent identity actually works today, an Agent only exists once a `tink-agent` process (with a container runtime) boots and connects to the Tink server, and it's tied to a `Hardware.spec.agentID` — which as far as I can tell isn't something the codebase supports today for a standing, hardware-independent Agent. Did you have something specific in mind?

That also raises a chicken-and-egg problem for our specific use case: the entire reason we need the webhook is that the target machine *can't* reach the Tinkerbell control plane yet — it's sitting on the wrong VLAN. So `abcd` can't be the target machine's own Agent (it doesn't exist yet). It would have to be a *separate*, independent, already-provisioned, permanently-running machine dedicated to executing this one Task.

If that's the intent, I think it's worth weighing that tradeoff explicitly: it trades your stated goal (avoid webhook code/auth living in Tinkerbell) for standing up and permanently operating a dedicated host (container runtime + `tink-agent`, enrolled as its own `Hardware` object, monitored, kept connected) purely to fire HTTP calls. That's arguably more operational surface than a webhook call made directly by the Tink Controller we already run — and it becomes a new single point of failure shared by every provisioning workflow that needs this gate, where today's design has no such dependency. Curious whether you see a lighter-weight version of this I might be missing.

**The polling Task isn't needed for our case**

Our switch reconfiguration call is synchronous — the API only returns once the VLAN move has actually been applied. There's no async gap to poll for, so the race condition Task 2 in your example is solving for doesn't apply here. `callWebhook` already blocks (with a timeout) and treats the call as failed/succeeded based on that single response — that's sufficient for a synchronous backend.

That said, today `callWebhook` only checks `resp.StatusCode` and discards the response body entirely (`webhook.go`, `callWebhook`). For extra safety on gating calls like this, we could extend `WebhookAction` with an optional way to assert on the response body (e.g. an expected substring or JSONPath match), not just the status code. Happy to add that to this PR if useful.

**v1alpha1 vs. v1alpha2**

You framed this as "I don't see a way to get this into the v1alpha1 spec" — I'd actually pose the question slightly differently: this PR is deliberately designed to be additive at the YAML level for existing pure-BMC `preparingActions`/`postActions` (a `- powerAction: off` entry parses exactly as it does today); the only real compatibility costs are documented in the PR description (batch job naming, one-time post-actions re-run on upgrade), and they're the necessary price of adding this to a spec with real deployed users.

The more interesting question, to me, is whether this design could go into v1alpha2 instead. Structurally, yes — `Customboot.PreOperations`/`PostOperations` there is the exact same list-of-BMC-operations shape, and the identical technique (retype the element to a union that embeds `bmc.Operations` inline plus an optional webhook entry) would apply directly, with *zero* compatibility cost since v1alpha2 has no production users yet. But v1alpha2 has no reconciler/execution engine at all right now — no template rendering, no BMC job orchestration, nothing that runs a Workflow. Schema compatibility isn't the blocker; the entire v1alpha2 controller doesn't exist yet, and that's a large, unscheduled, unrelated effort. So my instinct is this shouldn't block landing the feature in v1alpha1 now — but I'd be glad to carry this same interleaving design over to v1alpha2's `Customboot` once it has a working controller behind it. Curious whether you see it differently, or whether there's a middle ground I'm not seeing.

mcanevet:
Following up on the "abcd" standing-agent idea — I think I now understand the actual tradeoff precisely, want to confirm.

In v1alpha1 today, with bootOptions/customboot doing the reboot: since the whole Workflow sits in PREPARING (and no Task is served to any Agent) until the bmc.Job completes, a webhook Task on a standing agent can only run after rufio has already ordered power-on — not before it. So it's racing the physical machine's own boot time (BIOS/UEFI POST, firmware init, etc.) rather than strictly preceding it. Empirically our hardware takes 2-5 minutes to reach its DHCP client, which is a comfortable margin against the few seconds a reconcile-flip + agent poll + webhook call take — so this can work fine in v1alpha1 today, it's just relying on that measured hardware timing, not a structural guarantee.

The real, non-negligible advantage of v1alpha2's per-Task bootOptions is that it would let the webhook Task run before the bmc.Job/reboot is even triggered, making "webhook, then reboot" a guarantee by construction instead of something that happens to work because of how long our servers take to POST.

Is that a fair summary, or am I missing something?

mcanevet:
Looks like it work as I understand. So I have everything I want even with `v1alpha1`. @jacobweinstock I don't know what to do with this PR, but we can close it if you want or continue discussing about it.

PR #897, opened 2026-08-01
fix(smee): preserve tink-server hostname and add syslog FQDN flag for iPXE scripts

### LINKED ISSUE #531: --ipxe-script-tink-server-addr-port hostname is ignored, always replaced with --public-ipv4

### LINKED ISSUE #533: add --ipxe-script-syslog-fqdn flag for iPXE scripts

gianarb:
## Description

Lets operators use a resolvable **hostname/FQDN instead of a hard IP** in the iPXE scripts Smee generates. Needed for global bare-metal provisioning: split-horizon DNS, TLS certificate validation (certs are issued for hostnames), and reverse-proxy routing.

This combines and supersedes #532 and #534 (whose original contributor moved on). Those two PRs overlapped because #532's branch stacked the syslog commit on top of the tink-server fix, so #534's change appeared duplicated inside #532. They are two distinct concerns and are both addressed here.

### 1. Preserve tink-server hostname (Fixes #531)

`Convert()` in `cmd/tinkerbell/flag/smee.go` discarded the host from `--ipxe-script-tink-server-addr-port` and always substituted `--public-ipv4`, so `grpc_authority` could never be a hostname. It now preserves a user-provided host and only falls back to `publicIP` when the host is empty.

The computation was also moved **above** the `publicIP` early-return guard, so a hostname is preserved even when `--public-ipv4` is unset (an edge case not handled by the original PR).

### 2. Add `--ipxe-script-syslog-fqdn` (Fixes #533)

New flag to set `syslog_host` in iPXE scripts to a hostname instead of the IP from `--dhcp-syslog-ip`. Falls back to `--dhcp-syslog-ip` when unset, so behavior is unchanged for existing users.

## How Has This Been Tested?

- `go build ./...` passes.
- `go test ./cmd/tinkerbell/flag/... ./smee/...` passes.
- New table tests:
  - `Convert()` hostname/IP preservation, default-port, publicIP fallback, and hostname-with-unspecified-publicIP (uses the current 4-arg `Convert` signature).
  - `Config.syslogHost()` FQDN precedence and IP fallback.

## How are existing users impacted?

No impact — fully backward compatible. When neither new behavior is triggered (no hostname given / `--ipxe-script-syslog-fqdn` unset), output is identical to before.

## Checklist

- [x] added unit tests
- [x] backward compatible

### DIFF
--- a/smee/smee.go
+++ b/smee/smee.go
@@ -174,6 +174,9 @@ type IPXEHTTPScriptServer struct {
 	ExtraKernelArgs []string
 	KernelName      string
 	InitrdName      string
+	// SyslogFQDN is the hostname/FQDN for the syslog server in iPXE scripts.
+	// If empty, it falls back to DHCP.SyslogIP.
+	SyslogFQDN string
 }
 
 type DHCP struct {
@@ -373,7 +376,7 @@ func (c *Config) ScriptHandler(log logr.Logger) http.Handler {
 		Backend:               c.Backend,
 		OSIEURL:               c.IPXE.HTTPScriptServer.OSIEURL.String(),
 		ExtraKernelParams:     c.IPXE.HTTPScriptServer.ExtraKernelArgs,
-		PublicSyslogFQDN:      c.DHCP.SyslogIP.String(),
+		PublicSyslogFQDN:      c.syslogHost(),
 		TinkServerTLS:         c.TinkServer.UseTLS,
 		TinkServerInsecureTLS: c.TinkServer.InsecureTLS,
 		TinkServerGRPCAddr:    c.TinkServer.AddrPort,
@@ -384,6 +387,16 @@ func (c *Config) ScriptHandler(log logr.Logger) http.Handler {
 		InitrdName:            c.IPXE.HTTPScriptServer.InitrdName,
 	}
 	return jh.HandlerFunc()
+}
+
+// syslogHost returns the host used for the syslog_host kernel parameter in iPXE scripts.
+// It prefers the configured SyslogFQDN (a hostname/FQDN) and falls back to the DHCP syslog IP
+// when no FQDN is set.
+func (c *Config) syslogHost() string {
+	if c.IPXE.HTTPScriptServer.SyslogFQDN != "" {
+		return c.IPXE.HTTPScriptServer.SyslogFQDN
+	}
+	return c.DHCP.SyslogIP.String()
 }
 
 // ISOHandler returns an http.Handler that serves patched ISO images.
--- a/smee/smee.go
+++ b/smee/smee.go
@@ -411,7 +411,7 @@ func (c *Config) ISOHandler(log logr.Logger) (http.Handler, error) {
 		Patch: iso.Patch{
 			KernelParams: iso.KernelParams{
 				ExtraParams:        c.IPXE.HTTPScriptServer.ExtraKernelArgs,
-				Syslog:             c.DHCP.SyslogIP.String(),
+				Syslog:             c.syslogHost(),
 				TinkServerTLS:      c.TinkServer.UseTLS,
 				TinkServerGRPCAddr: c.TinkServer.AddrPort,
 			},
--- a/cmd/tinkerbell/flag/smee.go
+++ b/cmd/tinkerbell/flag/smee.go
@@ -2,6 +2,7 @@ package flag
 
 import (
 	"fmt"
+	"net"
 	"net/netip"
 	"strconv"
 	"strings"
@@ -106,6 +107,7 @@ func RegisterSmeeFlags(fs *Set, sc *SmeeConfig) {
 	fs.Register(IPXEHTTPScriptRetries, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.Retries, sc.Config.IPXE.HTTPScriptServer.Retries))
 	fs.Register(IPXEHTTPScriptRetryDelay, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.RetryDelay, sc.Config.IPXE.HTTPScriptServer.RetryDelay))
 	fs.Register(IPXEHTTPScriptOSIEURL, &url.URL{URL: sc.Config.IPXE.HTTPScriptServer.OSIEURL})
+	fs.Register(IPXEScriptSyslogFQDN, ffval.NewValueDefault(&sc.Config.IPXE.HTTPScriptServer.SyslogFQDN, sc.Config.IPXE.HTTPScriptServer.SyslogFQDN))
 	fs.Register(IPXEBinaryInjectMacAddrFormat, &ffval.Enum[constant.MACFormat]{
 		ParseFunc: macAddrFormatParser,
 		Valid:     []constant.MACFormat{constant.MacAddrFormatColon, constant.MacAddrFormatDot, constant.MacAddrFormatDash, constant.MacAddrFormatNoDelimiter},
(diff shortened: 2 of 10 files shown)


gianarb:
I fixed the bug "IPv6 bracket handling and empty-host case" exposed directly by my swap of args validation (my attempt was to highlight the invalid value earlier instead of silently moving over. 

The other one copilot got can be fixed in its own pr I think since it is as well there from before my changes. let me know

smee/smee.go
```diff
@@ -374,7 +377,7 @@ func (c *Config) ScriptHandler(log logr.Logger) http.Handler {
 		Backend:               c.Backend,
 		OSIEURL:               c.IPXE.HTTPScriptServer.OSIEURL.String(),
 		ExtraKernelParams:     c.IPXE.HTTPScriptServer.ExtraKernelArgs,
-		PublicSyslogFQDN:      c.DHCP.SyslogIP.String(),
+		PublicSyslogFQDN:      c.syslogHost(),
```
jacobweinstock, line 379: Hey @gianarb. The default iPXE script sets the syslog in two places. 
1. iPXE syslog server: https://github.com/tinkerbell/tinkerbell/blob/cc10c6f2655b622b88435e5cf2f45553662f0df4/smee/internal/ipxe/script/hook.go#L7
2. Kernel parameters: https://github.com/tinkerbell/tinkerbell/blob/cc10c6f2655b622b88435e5cf2f45553662f0df4/smee/internal/ipxe/script/hook.go#L25

The first one, `set syslog` requires that the value be an IPv4 address. [Reference](https://ipxe.org/cfg/syslog), while the second one can be a FQDN. 

One idea could be to do some iPXE script logic, see my suggestion here:
https://github.com/tinkerbell/tinkerbell/pull/534#issuecomment-3797839120)

gianarb:
I checked the ipxe binary and it contains the nslookup command so we are good. I also tried with Qemu and the change should be retro compatible :+1: 

I did not have the possibility to try this e2e yet

gianarb:
I tried this in playground I think it works as intended @jacobweinstock

jacobweinstock:
Thanks, @gianarb. Mind setting up your [GitHub verified signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification), please? Thanks!

jacobweinstock:
Hey @gianarb , thanks for this PR. A few concerns to discuss.

### 1. ISO boot path ignores the new syslog FQDN

`ScriptHandler` now resolves the syslog host via `c.syslogHost()`, but [`ISOHandler` still hardcodes the DHCP IP](https://github.com/tinkerbell/tinkerbell/blob/40e21405214ca09415c5ecb82270517914be534b/smee/smee.go#L414):

```go
// smee/smee.go — ISOHandler
Syslog: c.DHCP.SyslogIP.String(),
```

So an iPXE-booted machine gets `syslog_host=<fqdn>` while an ISO-booted machine gets `syslog_host=<ip>` for the same kernel arg.

#### Requested change
When the FQDN is specified it needs to work the same way across all uses, and specifically the ISO path.

### 2. New flag isn't wired into the Helm chart

`--dhcp-syslog-ip` is exposed as [`dhcpSyslogIP` in `values.yaml`](https://github.com/tinkerbell/tinkerbell/blob/40e21405214ca09415c5ecb82270517914be534b/helm/tinkerbell/values.yaml#L75) and [`TINKERBELL_DHCP_SYSLOG_IP` in `deployment.yaml`](https://github.com/tinkerbell/tinkerbell/blob/40e21405214ca09415c5ecb82270517914be534b/helm/tinkerbell/templates/deployment.yaml#L142-L143), but the new [`--ipxe-script-syslog-fqdn` flag](https://github.com/tinkerbell/tinkerbell/blob/40e21405214ca09415c5ecb82270517914be534b/cmd/tinkerbell/flag/smee.go#L386-L389) has no corresponding `values.yaml` entry or `TINKERBELL_IPXE_SCRIPT_SYSLOG_FQDN` env var.

#### Requested change
Add this in this PR or defer this to a follow up PR, either is fine.

### 3. `nslookup` now runs even when the syslog host is an IP

Previously the template did `set syslog <ip>` directly and unconditionally. Now every boot, including the common IP-only case, runs:

```
nslookup syslogserver <host>
set syslog ${syslogserver}
```

Two concerns:

- If `nslookup` fails, `${syslogserver}` is empty and `set syslog` silently sets nothing, with no error handling (`||`). The old IP path couldn't fail this way.
- This relies on iPXE's `nslookup` resolving a numeric IP literal without a reachable DNS server. That's generally true (iPXE has a numeric resolver), but it's now a hard dependency for **all** users, not just FQDN users.

#### Requested change
Whether the syslog get sets properly or not, it should not block the rest of the iPXE script.

Here's a possible solution:

```
# iPXE can only set the syslog server to an IP address, not a hostname (https://ipxe.org/cfg/syslog).
# If target is an IP, save it directly; if not, resolve it via nslookup directly into the syslog variable.
set check:ipv4 {{ .SyslogHost }} && set syslog {{ .SyslogHost }} || nslookup syslog {{ .SyslogHost }} || echo [WARN] Failed to resolve syslog host {{ .SyslogHost }}
clear check
```

jacobweinstock (approved):
Thanks, @gianarb !

PR #932, opened 2026-08-18
client-go: add Hardware client proof of concept

eddiezane:
## Summary

Propose a supported `github.com/tinkerbell/tinkerbell/client-go` nested module and add a one-resource proof of concept for `tinkerbell.org/v1alpha1` Hardware.

The proof includes:

- aggregate scheme registration
- conventional clientset constructors and discovery
- a JSON-negotiated, namespaced typed Hardware client with CRUD, list/watch, patch, and status behavior
- fake clientset
- typed lister
- shared informer factory with namespace and `TweakListOptions` support
- `ForResource`, `Start`, `Shutdown`, and `WaitForCacheSync`
- behavior-focused tests for REST paths, verbs, serialization, shared transport, fake behavior, and informer/lister behavior

`docs/client-go-design.md` inventories every root resource in v1alpha1 and v1alpha2 and describes the release and downstream-integration plan.

## Why this is a draft

This is intentionally limited to Hardware until maintainers agree on the two architectural questions below. It is a working design artifact, not yet the complete supported client surface.

1. Is `client-go/` with module path `github.com/tinkerbell/tinkerbell/client-go` the desired placement?
2. How should code generation handle the existing public `version/group` API layout when Kubernetes generators expect `group/version`?

The design document presents three options: codegen adapter aliases if a generator spike proves them, moving canonical definitions while retaining compatibility aliases, or a handwritten `gentype`-based transitional implementation. This POC uses the third option and does not relocate API packages.

## Compatibility and scope

- Existing `github.com/tinkerbell/tinkerbell/api/...` imports are unchanged.
- `api/go.mod` remains free of `k8s.io/client-go`.
- Kubernetes dependencies align with current repository versions (`v0.36.3`).
- The local API `replace` is for testing against this checkout and would be replaced by the matching `api/vX.Y.Z` requirement before publishing.
- Lease locking, ownership labels, feature flags, and workflow-state policy remain downstream concerns.

## Validation

- `go test -race ./...` in `client-go/`
- `go mod tidy -diff` in `client-go/` and `api/`
- `go test ./...` in `api/`
- `make lint`
- confirmed `api/go.mod` contains no `k8s.io/client-go` dependency

## Follow-up after design agreement

Expand the selected versions to every root resource; add any missing scheme registration; settle deterministic generation; integrate client-go testing, release tags, dependency updates, and documentation; then validate a temporary downstream migration that removes the handwritten `uds-remote-agent` clientset and informer plumbing.


eddiezane:
@jacobweinstock we talked about this on a community call at one point. There are some standard types/client k8s patterns that tink can generate and export for downstream consumers.

We'll swing by a call to chat this through and shape.

s-urbaniak:
Building on Eddie’s point about exporting the standard Kubernetes client patterns, an official `client-go` module would remove a significant amount of downstream plumbing.

There is also a separate API-level problem that typed clients, listers, and informers cannot solve by themselves: one physical-machine operation spans several independent resources.

A provisioning request is represented by a Tinkerbell `Workflow` referencing `Hardware`, while BMC actions are Rufio `Job`s referencing a `Machine`, and the association between them is the mutable `Hardware.spec.bmcRef`. There is therefore no single durable, atomically updated record answering “which operation currently owns this physical machine?”

Downstream consumers consequently have to:

- Correlate active `Workflow`s and `Job`s across resource types.
- Infer a Workflow’s Machine through a mutable BMC reference.
- Prevent Link/Unlink from invalidating that relationship while work is active.
- Avoid admitting conflicting operations while informer caches are still catching up with newly created resources.
- Recover coordination after restarts and serialize it across replicas.
- Make imperative operations such as reboot idempotent when reconciliation is retried.

Would there be interest in a higher-level Tinkerbell operation/orchestration CRD? A rough shape might be a `HardwareOperation` containing immutable Hardware and Machine targets, the requested action, and controller-owned status:

```yaml
apiVersion: tinkerbell.org/v1alpha1
kind: HardwareOperation
metadata:
  # Retrying the request with the same name adopts the existing intent.
  name: provision-server-1-request-1234
  namespace: default
spec:
  # These references remain immutable for the operation's lifetime.
  hardwareRef:
    name: server-1
  machineRef:
    name: bmc-1

  action:
    type: Provision
    provision:
      templateRef:
        name: ubuntu
      reboot: true
      bootMode: pxe

status:
  observedGeneration: 1
  phase: Running
  executionRef:
    apiVersion: tinkerbell.org/v1alpha1
    kind: Workflow
    name: provision-server-1-request-1234
  conditions:
    - type: Accepted
      status: "True"
      reason: TargetAvailable
    - type: Complete
      status: "False"
      reason: WorkflowRunning
```

Other action variants could represent PowerOn, PowerOff, Reboot, LinkBMC, or UnlinkBMC while retaining the same lifecycle and ownership model.

Its controller could persist intent before performing side effects, serialize operations by their affected Hardware or Machine, leave conflicting requests pending, and create or adopt deterministic child resources. After a restart, the controller could reconstruct everything from the CRDs and continue reconciliation.

The existing `Workflow` and Rufio `Job` APIs would remain the execution primitives; this would provide the missing common admission, ownership, and lifecycle layer above them. Even a smaller shared per-Hardware/Machine claim primitive would help downstream consumers avoid independently rebuilding this coordination logic.

I do not think this needs to block the client-go proposal, but it may be worth treating as a related API-design problem rather than leaving every downstream integration to invent its own locking and cross-resource state machine.

jacobweinstock:
Hey @eddiezane. Thanks for opening this.

A few points of discussion.
1. I like to see if we can get an idea if the broader community is interested in this. Maybe post in the Tinkerbell slack channel? Other ideas?
1. This is a significant increase in code that will need to be maintained.
1. For something like this I would lean towards it being in its own repo too.  This would open up opportunities for others to have maintainer status of just this repo without affecting the main tinkerbell/tinkerbell. 
1. The v1alpha2 API implementation is coming soon, should something like this wait for that?
1. Tinkerbell and CAPT currently use controller-runtime for most, if not all, of these primitives. This seems to be a big diversion from that. I'd like to more fully understand the use-cases and motivations, along with seeing if there are others in the community that have the same needs.

jacobweinstock:
Hey @s-urbaniak . Thanks for posting this idea. 

1. "_There is therefore no single durable, atomically updated record answering “which operation currently owns this physical machine?”_" 
   From a top level point of view, we can say there are only 2 types of operations: Workflows and BMC Jobs. Workflows current allow BMC Jobs to be defined within their spec. The BMC Job and Tasks created are correlated via things like owner refs and resource IDs. Also, we're talking about only 2 "operations". This is 2 calls to the kube api server to answer the question of ownership. So there are existing ways to answer this question. So that leads me to think that I'm not sure I'm following what the challenge is you're seeing.

1. _Infer a Workflow’s Machine through a mutable BMC reference. Prevent Link/Unlink from invalidating that relationship while work is active._
   I'm not familiar with this use case. How often are BMC references changing? Why are they changing? Who is changing them?
   
 1.  _Make imperative operations such as reboot idempotent when reconciliation is retried._
    There has been discussion in the past about declarative BMC operations. I like the idea, we need a proper design/proposal to discuss. I apologize, this use-case is unfamiliar to me as well. Reconciliation of what? BMC Jobs and Workflows are by design imperative operations. I don't think i understand the circumstance by which a reboot, for example, would get retried.
    
There has been discussion in the past about having a controller that basically does what CAPT does but for generic Workflows. A lifecycle controller of sorts. The challenge was we had lots of interesting ideas and no real use-cases. If I understand correctly, you're example CRD models this idea. We'd need to make sure we're not just creating a Workflow Spec with fields moved around. Also, the v1alpha2 API spec should probably be the target and this spec changes the lens in which the proposal is viewed.

Again, thanks for posted about this. I'm definitely open to further discussing this. A proper GitHub discussion might be an ideal location to continue the conversation.

s-urbaniak:
@jacobweinstock 
> From a top level point of view, we can say there are only 2 types of operations: Workflows and BMC Jobs. Workflows current allow BMC Jobs to be defined within their spec. The BMC Job and Tasks created are correlated via things like owner refs and resource IDs. Also, we're talking about only 2 "operations". This is 2 calls to the kube api server to answer the question of ownership. So there are existing ways to answer this question.

Yes we are on the same page 👍 The issue is not discovering ownership with the two API calls; it is making admission atomic: i.e.
```
if noActiveWorkflows(node1) && noActiveBMCJobs(node1) {
    createOperation(node1)
}
```
Two actors can concurrently observe no active operation and then both create conflicting operations for node1. We currently serialize this using a process-local mutually exclusive "reservations" concept but that does not protect against multiple replicas, or restarts.

What we effectively propose is an atomic, per-Hardware "claim" covering every operation on that machine: provisioning through a Workflow, as well as BMC-only operations such as power on/off through a Rufio Job.

We could implement this as a cooperative Kubernetes Lease per Hardware, using create/resource-version updates for atomic acquisition: https://kubernetes.io/docs/concepts/architecture/leases/. The broader question is whether this machine-operation admission and lifecycle coordination would also be useful as a first-class Tinkerbell resource, rather than every client implementing it independently.

s-urbaniak:
> I'm not familiar with this use case. How often are BMC references changing? Why are they changing? Who is changing them?

This is expected to happen rarely. It is an explicit administrative convenience/repair operation, not part of normal provisioning. Our UI lets a user associate a Tinkerbell Hardware resource with a Rufio Machine, or remove that association.

For example, when the original association was incorrect or the BMC was replaced. Our UI handles that request by patching `Hardware.spec.bmcRef`.

s-urbaniak:
> There has been discussion in the past about declarative BMC operations. I like the idea, we need a proper design/proposal to discuss. I apologize, this use-case is unfamiliar to me as well. Reconciliation of what? BMC Jobs and Workflows are by design imperative operations. I don't think i understand the circumstance by which a reboot, for example, would get retried.

Sorry, my bad. By reconciliation, I meant reconciliation of the proposed `HardwareOperation`, not reconciliation of an existing Rufio `Job`.

For example, its controller could create a reboot `Job`, crash before updating `HardwareOperation.status`, and then reconcile again. A deterministic child name derived from the `HardwareOperation` name or UID lets it adopt the existing `Job` instead of executing another reboot.

The `HardwareOperation` itself would need a stable, caller-provided request name so retries create or adopt the same operation.

s-urbaniak:
> Again, thanks for posted about this. I'm definitely open to further discussing this. A proper GitHub discussion might be an ideal location to continue the conversation.

No, thank you so much for considering these thoughts! 🙇

As discussed above, the above use cases could be solved using other primitives, such as Kubernetes Leases or our own higher-level custom resource. We mainly wanted to understand whether a shared higher-level abstraction would also be useful to the project. 👍

It is also completely reasonable for Tinkerbell to define its current resources as execution primitives and leave cross-resource orchestration to downstream consumers.

PR #954, opened 2026-08-28
smee: answer the doubled-slash form of the RPi config.txt request

### LINKED ISSUE #953: smee: Raspberry Pi netboot route answers only one of the two config.txt requests the bootloader makes
### Summary

`RPiNetbootRoute` in `smee/internal/ipxe/binary/route_rpi.go` matches the two inline files with an exact `switch req.Filename`:

```go
switch req.Filename {
case rpi.SerialNum + "/config.txt":
case rpi.SerialNum + "/cmdline.txt":
}
```

The Raspberry Pi bootloader asks for each of these **twice**: once as `<serial>/config.txt` and again as `<serial>//config.txt`, with a doubled slash. Only the first form matches, and it is the second that the firmware acts on. When it misses, the Pi discards the config it was just served, probes the default kernel names, finds nothing, and loops.

Real files are unaffected, because the rewritten path goes through `os.OpenRoot`, which resolves the doubled slash — `<serial>//overlays/overlay_map.dtb` is served correctly. So the two inline templates are the only casualties, which makes this present as a bad `config.txt` rather than a routing bug.

### Version

v0.25.0

### Expected Behaviour

Both request forms are the same request, and both should be answered from `netboot.rpi.configTxt` and `netboot.osie.kernelParams`.

### Actual Behaviour

The doubled-slash request falls through this route (`handled=false`), the router tries the remaining routes, and the client gets a 404 for a file it has just been told exists.

The failure is close to invisible. The only trace is at `V(1)`:

```
request path does not begin with SerialNum; skipping
```

and that line is not even reached — the prefix check passes, since `<serial>//config.txt` does start with `<serial>/`. The `switch` simply falls out the bottom, the path is rewritten to `<firmwarePath>//config.txt`, and `openAsset` misses because nothing wrote that file. With `logLevel: 0` there is nothing at all. What it looks like from the outside is a Pi that fetches `config.txt`, the dtb and `overlay_map.dtb`, then goes quiet for ~30s and restarts.

### Steps to Reproduce

1. A Raspberry Pi 5 flashed with `TFTP_PREFIX=0` (serial addressing), with `netboot.rpi.serialNum` / `firmwarePath` and `netboot.rpi.configTxt` + `netboot.osie.kernelParams` set on its Hardware, and `tftpAssetDir` populated with the kernel/initramfs/dtb/overlays.
2. Boot it and watch the TFTP requests at `smee.logLevel: 1`.

Observed: `serving RPI ConfigTxt` appears once, for `<serial>/config.txt`. The subsequent `<serial>//config.txt` produces no such line. The Pi then probes `kernel_2712.img`, `kernel8.img` and `kernel8_rt.img` — the defaults it uses when it has no usable `config.txt` — finds none, and restarts.

### Why the obvious workaround is a problem

The natural fix is to write `config.txt` and `cmdline.txt` into `tftpAssetDir` under `firmwarePath`, so the doubled-slash request finds something on disk. That works, but it defeats the purpose of the inline fields: those files are then **shared by every machine on a given `firmwarePath`**, so `worker_id` is fixed across all of them and only one machine per boot environment can exist at a time. Per-machine boot config is exactly what `netboot.rpi.configTxt` and `netboot.osie.kernelParams` are for, and the doubled-slash miss is what forces you back off them.

`firmwarePath` is documented as a shared directory for generic firmware ("avoid having to duplicate firmware files for each Raspberry Pi"), so giving each machine its own just to get a distinct `cmdline.txt` works against the field's design too.

### Possible fix

Compare on the cleaned path:

```go
suffix := req.Filename[len(rpi.SerialNum):]

switch path.Clean(suffix) {
case "/config.txt":
    ...
case "/cmdline.txt":
    ...
}

rewritten := rpi.FirmwarePath + suffix
```

One thing worth being deliberate about: only the **comparison** should be normalised, not the path handed to `openAsset`. `os.OpenRoot` already resolves the doubled slash for real files, so cleaning buys nothing there — but it costs something. `OpenRoot` refuses a path that escapes the root, and `<firmwarePath>/../../../etc/passwd` does; `path.Clean` rewrites that to `/etc/passwd` first, making the rewritten path `<firmwarePath>/etc/passwd`, which is inside the root and served if it happens to exist. Cleaning the rewrite path would turn a refusal into a hit.

I have this running with tests covering both slash forms for the inline files, the doubled slash on a rewritten asset, and the escape-vs-clean case above. Happy to open a PR if the approach looks right.

alebsack:
Fixes #953.

### The bug

`RPiNetbootRoute` matched the two inline files with an exact `switch req.Filename`. The Raspberry Pi 5 bootloader asks for each of them **twice** — once as `<serial>/config.txt` and again as `<serial>//config.txt`, with a doubled slash — so only the first form was answered.

This is observed Pi 5 behaviour, from one Pi 5 booting a custom OSIE. No other model has been tested and Raspberry Pi documents nothing of the kind, so the claim is deliberately scoped to the Pi 5. The fix does not depend on which models do it: `path.Clean` on the comparison is a no-op for a well-formed path, so a client that only ever sends the single-slash form takes exactly the same branch it does today.

The second is the one the firmware acts on. On a miss the Pi discards the config it was just served, probes the default kernel names (`kernel_2712.img`, `kernel8.img`, `kernel8_rt.img`), finds nothing and loops.

Real files were unaffected the whole time, because the rewritten path goes through `os.OpenRoot`, which resolves the doubled slash — `<serial>//overlays/overlay_map.dtb` was served correctly. That is why only `configTxt` and `kernelParams` broke, and why it presents as a bad `config.txt` rather than a routing bug.

It is also close to silent. The prefix check passes (`<serial>//config.txt` does start with `<serial>/`), so not even the `V(1)` "request path does not begin with SerialNum" line fires — the `switch` just falls out the bottom. From outside it looks like a Pi that fetched a few files and went quiet.

### The fix

Compare on `path.Clean` of the suffix, so both forms reach the inline templates.

**Only the comparison is normalised — the rewritten path keeps the raw suffix**, and that is deliberate. `OpenRoot` already resolves the doubled slash for real files, so cleaning buys nothing there, but it costs something: `OpenRoot` refuses a path that escapes the root, and `<FirmwarePath>/../../../etc/passwd` does. `path.Clean` rewrites that to `/etc/passwd` first, making the rewritten path `<FirmwarePath>/etc/passwd` — inside the root, and served if it exists. Cleaning the rewrite path would turn a refusal into a hit.

There is a test for exactly that case, with a decoy file in place so it is a real distinction rather than a coincidental miss. I checked it fails (`handled=true want=false`) when the rewrite path is cleaned.

### Tests

Four cases added to `TestRPiNetbootRoute`:

- `config.txt with a doubled slash served from ConfigTxt` — the bug
- `cmdline.txt with a doubled slash served from joined OSIE.KernelParams` — same, other file
- `doubled slash on a rewritten asset still served from disk` — pins the behaviour that already worked, so a future rewrite of the path handling does not lose it
- `escaping traversal is refused even when the cleaned path would resolve` — pins the raw-suffix decision above

### Also

Corrected a comment on `openAsset` while in there: it refuses paths that *leave* the root, not every `..`. One that resolves back inside is served — which is the distinction the fix depends on, so it seemed worth having the comment say so.

### Note

Independent of #952, which touches the same file. Whichever merges first, I will rebase the other.

### DIFF
--- a/smee/internal/ipxe/binary/route_rpi.go
+++ b/smee/internal/ipxe/binary/route_rpi.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"path"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -23,6 +24,9 @@ import (
 //     serves an inline value (RPI.ConfigTxt for config.txt; the joined
 //     OSIE.KernelParams for cmdline.txt) or rewrites the path's serial
 //     prefix to FirmwarePath and streams the file from AssetDir.
+//   - Matches the inline files on the cleaned path, so the doubled-slash
+//     form the bootloader also asks for ("<SerialNum>//config.txt") is
+//     answered as well as the single-slash one.
 //
 // Returns handled=false when there's no Hardware match, no RPI config on
 // the Hardware, no AssetDir, the path doesn't have the serial prefix, or
@@ -64,20 +68,45 @@ func (r RPiNetbootRoute) TryServe(ctx context.Context, req Request, w io.ReaderF
 		return false, nil
 	}
 
-	switch req.Filename {
-	case rpi.SerialNum + "/config.txt":
+	suffix := req.Filename[len(rpi.SerialNum):]
+
+	// Compared cleaned, because the bootloader asks for each of these TWICE:
+	// once as "<serial>/config.txt" and again as "<serial>//config.txt", with a
+	// doubled slash. Both are the same request. An exact match on req.Filename
+	// answers only the first, and it is the second that the firmware acts on --
+	// when it misses, the Pi discards the config it was just served, probes the
+	// default kernel names (kernel_2712.img, kernel8.img, kernel8_rt.img), finds
+	// nothing and loops forever. Observed on a Pi 5 booting a custom OSIE.
+	//
+	// The miss is close to invisible: this route simply returns handled=false,
+	// the router falls through, and the client gets a 404 for a file it has
+	// already been told exists. What it looks like from outside is a Pi that
+	// fetched a few files and then went quiet.
+	//
+	// Only the comparison is normalised. The rewritten path below keeps the raw
+	// suffix, for a reason worth spelling out: os.OpenRoot already resolves the
+	// doubled slash for real files, so cleaning buys nothing there -- but it
+	// costs something. OpenRoot refuses a path that escapes the root, and
+	// "<FirmwarePath>/../../../etc/passwd" does. Cleaning the suffix first
+	// rewrites that request to "/etc/passwd", making the rewritten path
+	// "<FirmwarePath>/etc/passwd", which is inside the root and served if it
+	// happens to exist. Normalising here would convert a refusal into a hit.
+	switch path.Clean(suffix) {
+	case "/config.txt":
 		log.Info("serving RPI ConfigTxt")
 		return serveTemplate(w, log, span, req.Filename, rpi.ConfigTxt)
-	case rpi.SerialNum + "/cmdline.txt":
+	case "/cmdline.txt":
 		cmdline := strings.Join(hw.OSIE.KernelParams, " ")
 		log.Info("serving cmdline.txt from OSIE.KernelParams", "params", hw.OSIE.KernelParams)
 		return serveTemplate(w, log, span, req.Filename, cmdline)
 	}
 
 	// The rewritten path mixes hardware-supplied FirmwarePath with the
-	// client-supplied suffix; openAsset confines it to AssetDir and rejects
-	// absolute paths and ".." traversal so neither can escape the directory.
-	rewritten := rpi.FirmwarePath + req.Filename[len(rpi.SerialNum):]
+	// client-supplied suffix; openAsset confines it to AssetDir via os.OpenRoot,
+	// so neither can escape the directory. Note OpenRoot refuses a path that
+	// leaves the root, not every ".." -- one that resolves back inside is served
+	// -- which is the distinction the comment above depends on.
+	rewritten := rpi.FirmwarePath + suffix
 	log.V(1).Info("attempting to load rewritten file from asset dir", "rewritten", rewritten, "assetDir", r.AssetDir)
 
 	file, err := openAsset(r.AssetDir, rewritten)
--- a/smee/internal/ipxe/binary/route_rpi.go
+++ b/smee/internal/ipxe/binary/route_rpi.go
@@ -70,13 +70,20 @@ func (r RPiNetbootRoute) TryServe(ctx context.Context, req Request, w io.ReaderF
 
 	suffix := req.Filename[len(rpi.SerialNum):]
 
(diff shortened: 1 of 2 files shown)


smee/internal/ipxe/binary/route_rpi.go
```diff
@@ -64,20 +68,45 @@ func (r RPiNetbootRoute) TryServe(ctx context.Context, req Request, w io.ReaderF
 		return false, nil
 	}
 
-	switch req.Filename {
-	case rpi.SerialNum + "/config.txt":
+	suffix := req.Filename[len(rpi.SerialNum):]
+
+	// Compared cleaned, because the bootloader asks for each of these TWICE:
```
jacobweinstock: Hey @alebsack. I dont believe that the double slash behavior occurs on anything but the Raspberry Pi 5. Mind making this comment clear that this because of observed behavior of the Raspberry PI 5 bootloader?

jacobweinstock (dismissed):
Thanks, @alebsack. One small comment request and we'll get this merged. Thanks for fixing this!

smee/internal/ipxe/binary/route_rpi.go
alebsack: Done in 5238f8e8 — and your read matches the evidence, so the comment was overstating it.

Everything behind this change comes from a single Pi 5 (serial prefix `714e7d7f`) booting a custom OSIE. No other model has been tested, so "the bootloader" was doing work the observation does not support. It now reads:

```go
// This is observed behaviour of the Pi 5 bootloader specifically, seen on a
// Pi 5 booting a custom OSIE. It has not been observed on other models, and
// no such behaviour is documented by Raspberry Pi. Cleaning the comparison
// is a no-op for a client that only ever sends the single-slash form, so
// the route behaves identically for models that do not do this.
```

I added that last sentence because the fix does not depend on which models do it: `path.Clean` on the comparison changes nothing for a well-formed path, so a Pi 4 that only ever sends the single-slash form takes exactly the same branch it does today.

Worth flagging that I could not find this documented anywhere on either side, which is partly why I would not want the comment claiming more than was seen. There is no report of the doubled slash in `raspberrypi/rpi-eeprom` or `raspberrypi/firmware`, and none in Tinkerbell. The closest thing is a maintainer note in [rpi-eeprom#68](https://github.com/raspberrypi/rpi-eeprom/issues/68) that the bootloader assembles TFTP paths by "simple string concatenation" with no significance to the forward slash — which is a plausible mechanism for an extra separator, but is from 2019 and about `TFTP_PREFIX_STR`, so I am not presenting it as confirmation.

I have since put a log line on our own build that records the requested filename exactly as it arrives off the wire, before `path.Base` collapses it, specifically to settle the Pi 4 question. It is deployed but not yet captured — the Pi 4 here points at a different TFTP server and needs a DHCP change first. If it turns out the Pi 4 does it too I will come back and correct the comment; if you would rather hold this PR until then, that is fine by me.

jacobweinstock (approved):
Thanks, @alebsack

PR #961, opened 2026-09-01
Add block device path to in-band inventory

mcanevet:
## Summary
- Adds a `Path` field (udev `ID_PATH`, e.g. `pci-0000:17:00.0-nvme-1` — the `/dev/disk/by-path` symlink name) to block devices reported by in-band inventory collection.
- Threads the field through the full pipeline: proto `Block` message (field 12), `data.Block` / `tinkerbell.BlockDevice`, the tink-server gRPC-to-CRD and Agent-to-UI converters, and a new "Path" column in the Hardware UI's in-band Block Devices table.
- In-band only — the Agent populates it; there's no out-of-band equivalent.

Path complements the existing identity fields rather than duplicating them: `Name` is kernel-assigned and can reorder across reboots, `WWN`/`SerialNumber` identify the drive itself, and `Path` identifies the slot/port the drive occupies — the key that survives drive swaps and can be joined against location-based (bay) naming in out-of-band inventory.

## Test plan
- [x] `go test ./...` passes (unit tests updated in `tink/server/internal/grpc`, `ui/templates`)
- [x] Verified rendered Hardware UI detail page shows the new Path column for in-band block devices

### DIFF
--- a/api/v1alpha1/tinkerbell/hardware.go
+++ b/api/v1alpha1/tinkerbell/hardware.go
@@ -625,6 +625,12 @@ type BlockDevice struct {
 	// sources report a drive, this is the key they can be joined on.
 	//+optional
 	WWN string `json:"wwn,omitempty"`
+	// Path is the device's position in the topology (udev ID_PATH, e.g.
+	// "pci-0000:17:00.0-nvme-1"), i.e. /dev/disk/by-path/<path>. Stable where Name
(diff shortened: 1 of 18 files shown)


mcanevet:
Rendered the Hardware detail page (In-Band Attributes → Block Devices) with fixture data to show the new Path column:

Block Devices table with Path column

jacobweinstock (approved):
@mcanevet, thanks for your patience on this review!

Issue #764, opened 2026-05-18
open telemetry end-to-end


jacobweinstock:
Tinkerbell already has some open telemetry for things like DHCP. It would be nice to plum open telemetry into a full Workflow. The Workflow being the initial entry point. This would capture everything in the Workflow, to the BMC operations, to the network or ISO booting, to the Workflow running, to the cloud-init calls, etc.

Some investigation should be made into Kubernetes events as well. This could be in conjunction with open telemetry or not but as they would potential be very similar in nature we should investigate the kubernetes event path. If another issue should be opened just for kubernetes events that is fine.

Another point of investigation should also be kubernetes conditions. Most if not all Tinkerbell CRDs should have conditions defined under the CRD status field. These are also similar in nature to both open telemetry and kubernetes events. If another issue should be opened just for conditions that is fine.

gianarb:
I feel pretty good working at this one 👍 I will looking into it. Are you used to assign issues to contributors? I constantly keep an eye on assigned issues and it helps me stay up to date 🥺

PR #962, opened 2026-09-01
smee: preserve traceparent sampling flags

riccardomenegazzo:
## Description

Smee currently extracts a W3C `traceparent` from iPXE/TFTP filenames so that provisioning requests can remain connected to an upstream distributed trace.

However, while the trace ID and parent span ID are preserved, the incoming `trace-flags` byte is ignored and the reconstructed `SpanContext` is always marked as sampled:

```go
TraceFlags: trace.FlagsSampled
```

This means that an upstream parent such as:

```text
00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-00
```

is incorrectly converted into a sampled remote parent inside Smee.

Besides changing the upstream sampling decision, this can create unnecessary telemetry and makes the propagated trace context semantically different from the one received by Smee.

This PR fixes that behavior by delegating traceparent parsing to OpenTelemetry's W3C `TraceContext` propagator instead of reconstructing the `SpanContext` manually.

### What changes

The traceparent embedded in the filename is now:

1. detected and separated from the original requested filename;
2. parsed using `propagation.TraceContext`;
3. accepted only when OpenTelemetry considers the W3C trace context valid;
4. inserted into the returned context as a remote parent without modifying its trace flags.

This preserves the trace context as received, including:

* `00` — not sampled
* `01` — sampled
* `02` — random trace ID
* `03` — sampled + random trace ID

The filename matching is also anchored to the end of the request so that filenames containing additional trailing data are not accidentally interpreted as valid traceparents.

### Why this matters

Tinkerbell already propagates tracing information across DHCP, PXE/TFTP, and provisioning components. Preserving the incoming trace context is important for maintaining correct end-to-end observability semantics.

A downstream service should not silently change the parent's sampling decision while propagating the same trace.

Using OpenTelemetry's own W3C parser also avoids maintaining a partial traceparent implementation in Smee and keeps parsing behavior aligned with the OpenTelemetry version already used by the project.

This contributes to the broader end-to-end OpenTelemetry work tracked in #764.

### Testing

Added regression coverage for:

* non-sampled traceparent (`00`);
* sampled traceparent (`01`);
* random trace ID flag (`02`);
* sampled + random flags (`03`);
* reserved flags in W3C traceparent version `00`;
* all-zero trace IDs;
* all-zero span IDs;
* ordinary filenames without trace context;
* traceparent-shaped suffixes followed by unexpected trailing data;
* preservation of the original context when the embedded traceparent is invalid.

### Compatibility

There are no API or configuration changes.

Valid existing sampled traceparents continue to behave as before. The behavioral change is limited to preserving the actual flags supplied by the upstream trace context rather than forcing every propagated trace to be sampled.

Fixes part of #764.

### DIFF
--- a/smee/internal/ipxe/binary/binary.go
+++ b/smee/internal/ipxe/binary/binary.go
@@ -17,6 +17,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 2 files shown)


jacobweinstock:
Hey @riccardomenegazzo . Thanks for this contribution. Mind setting up GitHub verified signatures please. Thanks!

Reference: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

riccardomenegazzo:
> Hey @riccardomenegazzo . Thanks for this contribution. Mind setting up GitHub verified signatures please. Thanks!
> 
> Reference: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Thank you so much for the info! I have just provided the verified signature

jacobweinstock (approved):
Thanks @riccardomenegazzo!

PR #983, opened 2026-09-10
ipxe: bump ipxe to 2.0.0+ @ ff6e520

rpardini:
- 🌿 ipxe/ipxe@ff6e52063e0b37062394fe37b9788af25175e7af
- 🐸 scripts: default GH repo to env `IPXE_TARGET_GH_OWNER_REPO` so job can run on forks too
  - our GHA workflow already set the env, but it was not used anywhere


rpardini:
Motivation for the bump:

1) Staying up-to-date 😉 
2) (some) Support for USB network cards.
2) iPXE 2.0.0 has initial Secure Boot support (which would need more effort, as it needs a `shim`, and we can't customize iPXE/would need to use the stock one/ implement `autoexec.ipxe` route); adopting it early-ish opens the path for exploration.

